PR #53, opened 2019-01-15
XML schema support

### LINKED ISSUE #48: XML Schemas
Using a format like the one used by Mondrian.

MarcioPorto:
Adds XML schema support.

Closes #48


hwchen:
I was thinking of implementing `From<SchemaConfigXML> for SchemaConfigJSON`, but your solution of simply serializing to Json is a lot less code :).

I think that for now this is a fine solution; serde is pretty fast, and it's a one-time cost. Have you done any benchmarks to see how much longer it takes to start up an xml schema (not to micro-optimize, I just want a general idea)?

One change I'd like to make is to make `JSON` -> `Json` and `XML` -> `XML` (this is a general convention for Rust).

hwchen:
great! Thanks :)

PR #54, opened 2019-01-15
Feature/sql gen

hwchen:
I've got the basics in, wanted to open up for comment. Still a bit messy and wip, but wanted to get comments on the interface.

- `Backend` trait now has `generate_sql`
- hooked up server to use `generate_sql`
- hooked default implementation of `generate_sql` to `core::standard_sql`
- moved all other sql generation to clickhouse backend.
- stuff like `DrilldownSql` etc have been moved to `query_ir` module, which also has a new `QueryIr` struct which is used to generate sql.

I care most about finalizing good naming and interface for the `QueryIr` query intermediate representation that's passed to `generate_sql`.

Other stuff can be more easily cleaned up


hwchen:
Going to merge as is, to be able to move on with development on other features. And so that anybody can experiment with this split for implementing new backends.

More refinements on the `Backend` trait, and renaming, are welcome in the future.

jspeis:
Awesome!

PR #55, opened 2019-01-15
Flush (using RwLock)

MarcioPorto:


tesseract-server/src/handlers/flush.rs
```diff
@@ -1,7 +1,65 @@
...
+    }
+    let query_res = QS_NON_STRICT.deserialize_str::<FlushQueryOpt>(&query);
+    let query = match query_res {
+        Ok(q) => q,
+        Err(err) => {
+            return Ok(HttpResponse::BadRequest().json(err.to_string()));
+        },
+    };
+
+    let db_secret = match env::var("TESSERACT_FLUSH_SECRET") {
```
hwchen: Let's move this into the `main` module, with the other config stuff. That will also prevent `env::var` from being called on every request.

tesseract-server/src/handlers/flush.rs
```diff
@@ -1,7 +1,65 @@
...
+            Ok(val) => val,
+            Err(err) => {
+                error!("{}", err);
+                return Ok(HttpResponse::InternalServerError().finish());
+            },
+        };
+
+        // Update shared schema
+        let mut w = req.state().schema.write().unwrap();
+        *w = schema.clone();
```
hwchen: Just wanted to double check here that a clone is required.

tesseract-server/src/handlers/flush.rs
hwchen: This might require holding the secret in `AppState`. Perhaps using `Option`, in case we don’t want to allow not flushing when there’s no secret?

hwchen:
Sorry, ended up merging the sql gen changes first and now there's some conflicts

tesseract-server/src/handlers/flush.rs
```diff
@@ -1,7 +1,61 @@
...
+    let db_secret = match &req.state().env_vars.flush_secret {
+        Some(db_secret) => db_secret,
+        None => { return Ok(HttpResponse::Unauthorized().finish()); }
+    };
+
+    if query.secret == *db_secret {
+        info!("Flush internal state");
+
+        // Read schema again
+        let schema = match schema_config::read_schema() {
```
hwchen: Same story here for reading the schema; I think that we can use the schema filepath which is now passed into state. For me, it's a bit more explicit, plus I think it will avoid reading the env var both outside and inside the `read_schema` fn (from what I see in `main`)

tesseract-server/src/app.rs
```diff
@@ -18,16 +18,24 @@ use crate::handlers::{
 use std::sync::{Arc, RwLock};
 
 
+#[derive(Debug, Clone)]
+pub struct EnvVars {
+    pub flush_secret: Option<String>,
+    pub database_url: String,
+    pub schema_filepath: Option<String>,
```
hwchen: Can we make this not an `Option`? I think we need the schema filepath no matter what. Or, are you preparing for eventually wanting to support the dynamic schema refresh from api? If so, I'd probably make the schema filepath a value in an enum `SchemaSource` or something, with the other value `schema_refresh_endpoint` or something. Let me know what you think.

tesseract-server/src/app.rs
```diff
@@ -9,21 +9,52 @@ use crate::db_config::Database;
...
+}
+
+/// Represents the location of a remote schema file.
+pub struct RemoteSchema {
+    endpoint: String,
+}
+
+/// Holds data about the source of a schema file.
+#[derive(Debug, Clone)]
+pub enum SchemaSource {
```
hwchen: the records `LocalSchema` and `RemoteSchema` only need to be declared inside the enum, the struct declarations are not needed in this case.

hwchen:
Looks great! Thanks, gonna merge now.

PR #71, opened 2019-01-29
Logic layer #1

### LINKED ISSUE #59: [logic layer] ability to retrieve largest/smallest year value
either through `Year=latest`/`Year=oldest` or `Year=max`/`Year=min`

### LINKED ISSUE #60: [logic layer] specify measures and drilldowns without giving an explicit cube

### LINKED ISSUE #83: [logic layer] basic api, one cube
- cube
- drilldown
- measure (just one for now)
- time
arbitrary cuts
sort

MarcioPorto:
closes #59 
closes #60 
closes #83


tesseract-core/src/schema.rs
```diff
@@ -91,6 +91,38 @@ pub struct Cube {
     pub measures: Vec<Measure>,
 }
 
+impl Cube {
+    /// Returns a Vec<String> of all the dimension name options for a given Cube.
+    pub fn get_all_dimension_names(&self) -> Vec<String> {
```
hwchen: It might be better to use `LevelName` from the module `names`, instead of `String`. I haven't looked at the rest of the code yet, but I would think it would make it easier to work with later, for matching/filtering on a dimension.

Or, even without guessing at future usage, it's nice to have something stronger than stringly typing, and I think that `LevelName` is appropriate here.

tesseract-core/src/schema.rs
```diff
@@ -91,6 +91,38 @@ pub struct Cube {
...
+                    dimension_names.push(format!("{}.{}.{}", dimension_name, hierarchy_name, level_name).to_string());
+                }
+            }
+        }
+
+        dimension_names
+    }
+
+    /// Returns a Vec<String> of all the measure names for a given Cube.
+    pub fn get_all_measure_names(&self) -> Vec<String> {
```
hwchen: Similarly, here I'd prefer to use `Measure` from the `names` module. It doesn't add much for the moment I guess, but it's consistent with using `LevelName` above

tesseract-core/src/schema.rs
hwchen: (Also, I think that this fn is getting all level names, instead of dimension names)

tesseract-server/src/app.rs
```diff
@@ -56,14 +63,32 @@ pub fn create_app(backend: Box<dyn Backend + Sync + Send>, db_type: Database, sc
...
+
+        // Aggregation
         .resource("/cubes/{cube}/aggregate", |r| {
             r.method(Method::GET).with(aggregate_default_handler)
         })
         .resource("/cubes/{cube}/aggregate.{format}", |r| {
             r.method(Method::GET).with(aggregate_handler)
         })
+
+        // Aggregation + Logic Layer
```
hwchen: I apologize, I may have not communicated clearly enough about the api here.

My plan was to have a `/logic-layer` route, and no `/cubes/{cube}/logic-layer` route. This `logic-layer` route will handle any query that requires inference, such as the cube resolution by measure, dimension resolution by cut, and member/dim resolution for time.

I'll keep reading further to see exactly what differs between the logic layer route and the aggregate route, and may add more comments here.

tesseract-server/src/handlers/aggregate.rs
```diff
@@ -58,7 +62,7 @@ pub fn do_aggregate(
         static ref QS_NON_STRICT: qs::Config = qs::Config::new(5, false);
     }
     let agg_query_res = QS_NON_STRICT.deserialize_str::<AggregateQueryOpt>(&query);
-    let agg_query = match agg_query_res {
+    let mut agg_query = match agg_query_res {
```
hwchen: let's remove the `mut` here, I don't think it's necessary anymore.

tesseract-server/src/handlers/aggregate.rs
```diff
@@ -120,11 +123,12 @@ pub fn do_aggregate(
         .responder()
 }
 
-#[derive(Debug, Deserialize, Serialize)]
+#[derive(Debug, Clone, Deserialize, Serialize)]
 pub struct AggregateQueryOpt {
-    drilldowns: Option<Vec<String>>,
-    cuts: Option<Vec<String>>,
-    measures: Option<Vec<String>>,
+    pub drilldowns: Option<Vec<String>>,
```
hwchen: I think all these pub can be removed too?

tesseract-server/src/handlers/logic_layer/aggregate.rs
```diff
@@ -0,0 +1,169 @@
...
+            );
+        },
+    };
+
+    finish_aggregation(req, agg_query, cube, format)
+}
+
+/// The last few aggregation operations are common across all different routes.
+/// This method implements that step to avoid duplication.
+pub fn finish_aggregation(
```
hwchen: If I'm understanding correctly, the year latest stuff is here. From what I mentioned above, I think we should put this with the other logic layer stuff (cube resolution endpoint)

tesseract-server/src/handlers/aggregate.rs
```diff
@@ -120,11 +123,12 @@ pub fn do_aggregate(
...
+    pub cuts: Option<Vec<String>>,
+    pub measures: Option<Vec<String>>,
+    pub year: Option<String>,
```
hwchen: If we're integrating the year latest stuff with logic layer, this will get moved to another endpoint.

tesseract-server/src/handlers/logic_layer/aggregate.rs
```diff
@@ -0,0 +1,169 @@
...
+    req: HttpRequest<AppState>,
+    mut agg_query: AggregateQueryOpt,
+    cube: String,
+    format: FormatType
+) -> FutureResponse<HttpResponse> {
+    // Process `year` param (latest/oldest)
+    match &agg_query.year {
```
hwchen: For the year api, it might be nice to limit the values we allow. So here I would try to take the `String` in `agg_query.year` and convert it into an enum using `TryFrom`.

I might do something like
```
pub enum Year {
  First,
  Last,
  Value(u32),
}
```
Let me know your thoughts on this one.

hwchen: Sorry I didn't think of this earlier, since this would create a lot of changes for you. I was a bit rushed before I left.

tesseract-server/src/handlers/logic_layer/detection.rs
```diff
@@ -0,0 +1,187 @@
...
+}
+
+/// Detects which cube to use based on the drilldowns, cuts and measures provided.
+/// In case the arguments are present in more than one cube, the first cube to match all
+/// requirements is returned.
+pub fn detect_cube(schema: Schema, agg_query: AggregateQueryOpt) -> Result<String, Error> {
+    let drilldowns = match agg_query.drilldowns {
+        Some(drilldowns) => {
+            let mut d: Vec<String> = vec![];
+            for drilldown in drilldowns {
```
hwchen: it looks like you're converting the querystring drilldown to the unique level name here? If so, can this be done by converting to `Drilldown` or `LevelName` from from the `names` module?

tesseract-server/src/handlers/logic_layer/detection.rs
```diff
@@ -0,0 +1,187 @@
...
+use log::*;
+use serde_derive::{Serialize, Deserialize};
+use serde_qs as qs;
+use std::convert::{TryFrom, TryInto};
+use tesseract_core::{Schema, Cube};
+use tesseract_core::format::{format_records, FormatType};
+use tesseract_core::Query as TsQuery;
+
+use crate::app::AppState;
+use crate::handlers::aggregate::AggregateQueryOpt;
```
hwchen: This route will need to use a different api than the other aggregate handler. Once again, sorry for not specifying better before.

This will be good to document as part of the pr also.

If you check canon's logic layer readme, you'll see that some of the keys in the query string are the same, but that they are all comma separated lists in a string instead of the cube api's list notation `[]`.

Also cuts are processed as key-value pairs, instead of as `cuts[]={member}`. This is where you'll have to use:
```
#[derive(Deserialize)]
struct LogicLayerQueryOpt {
  measure: String,
  year: Year,
  #[serde(flatten)]
  cuts: std::collections::HashMap,
}
```

And for now, it's fine to resolve to just one cube, but in the future we'll need to map multiple measures to their cube.

tesseract-server/src/handlers/logic_layer/detection.rs
```diff
@@ -0,0 +1,187 @@
...
+        None => vec![],
+    };
+
+    let measures = match agg_query.measures {
+        Some(measures) => measures,
+        None => vec![],
+    };
+
+    // TODO: Avoid clone here?
+    for cube in schema.cubes {
```
hwchen: We don't have to do this now, but the `iter().filter()` api might be nice to use.

tesseract-server/src/logic_layer/cache.rs
```diff
@@ -0,0 +1,196 @@
...
+#[derive(Debug, Clone)]
+pub struct CubeCache {
+    pub name: String,
+    pub time_dim: Dimension,
+    pub years: Vec<i16>,
+}
+
+impl CubeCache {
+    /// Returns dimension name in the format: `Dimension.Hierarchy.Level`.
+    pub fn get_time_dim_name(&self) -> String {
```
hwchen: Would prefer to use `LevelName` from `names` module.

tesseract-server/src/logic_layer/cache.rs
```diff
@@ -0,0 +1,196 @@
...
+        format!("{}.{}.{}",
+            self.time_dim.name,
+            self.time_dim.hierarchies[0].name,
+            self.time_dim.hierarchies[0].levels[0].name,
+        ).to_string()
+    }
+
+    pub fn get_year_cut(&self, s: String) -> Result<String, Error> {
```
hwchen: it's mentioned a little above. But I think it would be nice to have a `Year` enum, with `Latest` and `First` and `Value(usize)` so that it's really clear in the code what the allowed values are.

tesseract-server/src/app.rs
MarcioPorto: I assumed that we wanted to have the ability to specify which cube we want to query. One of the annoying things about the current logic layer, in my opinion, is that you can't do this. There are cases where two separate cubes have the same measure and dimension names. These cases were common in the Authority Health project. In those cases, it was impossible for the logic layer to differentiate between the different cubes so I had to change some names to things that don't make much sense. For example, I had to change a dimension name from `Sex` to `Diabetes Gender` so it could pick up that cube. This doesn't make much sense since gender is gender regardless of whether someone has diabetes. 

So personally I'd prefer to have a separate endpoint where the cube can be specified. As far as other logic layer functionality, everything after the cube detection step in the `/logic-layer` endpoint would be shared among both endpoints in the `finish_aggregation` method.

tesseract-server/src/handlers/logic_layer/aggregate.rs
MarcioPorto: Right now this is being used by both endpoints. Also, see my comment above about handling both endpoints.

tesseract-server/src/app.rs
hwchen: Thanks so much for bringing up this point. Let's discuss offline a bit.

tesseract-core/src/lib.rs
```diff
@@ -118,6 +118,80 @@ impl Schema {
...
+        // Here, we find the dimension and hierarchy names for the given level names.
+        // NOTE: Failing silently for queries with multiple drilldowns if not all of
+        //       the level names are found.
+        let cube_obj = self.cubes.iter()
+            .find(|c| &c.name == &cube)
+            .ok_or(format_err!("Could not find cube"))?;
+
+        let mut drilldowns: Vec<Drilldown> = vec![];
+        let mut cuts: Vec<Cut> = vec![];
+        let mut properties: Vec<Property> = vec![];
```
hwchen: Should we omit properties? At least I'm not sure how to specify properties, since they also need an associated level drilldown.

MarcioPorto: I talked to Dave and he told me the current logic layer requires only the property name and figures the rest out. So that's what I implemented here.

tesseract-core/src/lib.rs
```diff
@@ -119,6 +119,11 @@ impl Schema {
         if query.drilldowns.is_empty() && query.cuts.is_empty(){
             return Err(format_err!("Either a drilldown or cut is required"));
         }
+
```
hwchen: Is this a repeat of the check above?

tesseract-core/src/lib.rs
```diff
@@ -776,6 +781,12 @@ impl Schema {
 
         Ok(column)
     }
+
+    pub fn get_cube_by_name(&self, cube_name: &str) -> Result<&Cube, Error> {
```
hwchen: This is good, I think that I should probably refactor a bunch of methods on `Schema` to be on `Cube` instead. This will help move towards that.

tesseract-core/src/query_ir.rs
```diff
@@ -179,13 +179,17 @@ pub struct CutSql {
 impl CutSql {
     pub fn members_string(&self) -> String {
         let members = match self.member_type {
-            MemberType::NonText => join(&self.members, ", "),
+            MemberType::NonText => {
+                println!("{:?}", self);
```
hwchen: let's remove the print or change this to `debug!`

tesseract-core/src/schema.rs
```diff
@@ -138,6 +141,83 @@ pub struct Cube {
     pub annotations: Option<Vec<Annotation>>,
 }
 
+impl Cube {
```
hwchen: These cube methods have a good feel, and reflect how I think I'll be moving some `Schema` functionality to `Cube`

tesseract-core/src/sql.rs
```diff
@@ -33,9 +33,9 @@ pub(crate) fn standard_sql(
     // hack for now... remove later
     fn agg_sql_string(m: &MeasureSql) -> String {
         match &m.aggregator {
-            Aggregator::Sum => format!("sum"),
-            Aggregator::Count => format!("count"),
-            Aggregator::Average => format!("avg"),
+            Aggregator::Sum => format!("sum({})", &m.column),
```
hwchen: Thanks for the fixes here; I'm not working much with general sql yet, so this is very helpful

tesseract-server/src/logic_layer/cache.rs
```diff
@@ -0,0 +1,194 @@
...
+        let df = match sys.block_on(future) {
+            Ok(df) => df,
+            Err(err) => {
+                return Err(format_err!("Error populating cache with backend data: {}", err));
+            }
+        };
+
+        // TODO: Do we want to return an error if no columns are returned?
+        if df.columns.len() >= 1 {
+            let mut original_years = match &df.columns[0].column_data {
```
hwchen: Just a heads up that I'll be adding null support in the near future, which will add some more `ColumnData` types like `NullableInt8`, etc.

tesseract-server/src/schema_config.rs
```diff
@@ -15,7 +16,38 @@ pub fn read_schema(schema_path: &String) -> Result<Schema, Error> {
     } else if schema_path.ends_with("json") {
         schema = Schema::from_json(&schema_str)?;
     } else {
-        panic!("Schema format not supported");
+        return Err(format_err!("Schema format not supported"))
+    }
+
+    // Check each cube for unique level and property names
```
hwchen: Thanks for doing this check :) We should make sure to document this.

tesseract-server/src/schema_config.rs
```diff
@@ -15,7 +16,38 @@ pub fn read_schema(schema_path: &String) -> Result<Schema, Error> {
...
+    for cube in schema.cubes.clone() {
+        let mut levels = HashSet::new();
+        let mut properties = HashSet::new();
+
+        for dimension in cube.dimensions.clone() {
+            for hierarchy in dimension.hierarchies.clone() {
+                for level in hierarchy.levels.clone() {
+                    if levels.contains(&level.name) {
```
hwchen: https://doc.rust-lang.org/1.0.0/std/collections/struct.HashSet.html#method.insert

I think that insert allows you to do this with one fn call, because it returns `false` if value is already in the set. (same goes for the `properties` check

tesseract-core/src/names.rs
```diff
@@ -24,6 +24,8 @@ use failure::{Error, bail, format_err, ensure};
 use serde_derive::{Deserialize, Serialize};
 use std::fmt;
 use std::str::FromStr;
+use std::collections::HashMap;
```
hwchen: remove? I don't think it's used anymore.

tesseract-core/src/query.rs
```diff
@@ -1,5 +1,6 @@
 use failure::{Error, format_err, bail};
 use std::str::FromStr;
+use std::collections::HashMap;
```
hwchen: remove? I don't think it's used anymore

tesseract-server/Cargo.toml
```diff
@@ -17,6 +17,7 @@ pretty_env_logger = "0.2.4"
 serde = "1.0.70"
 serde_derive = "1.0.70"
 serde_qs = "0.4.1"
+serde_urlencoded = "0.5.1"
```
hwchen: I thought serde_qs will also do the deserialization of query strings? (which is what I think you're using this for)

tesseract-server/src/app.rs
```diff
@@ -59,20 +64,32 @@ pub fn create_app(backend: Box<dyn Backend + Sync + Send>, db_type: Database, sc
...
+        // Aggregation
         .resource("/cubes/{cube}/aggregate", |r| {
             r.method(Method::GET).with(aggregate_default_handler)
         })
         .resource("/cubes/{cube}/aggregate.{format}", |r| {
             r.method(Method::GET).with(aggregate_handler)
         })
+
+        // Logic Layer
+        .resource("/logic-layer", |r| {
```
hwchen: I realized after looking at logic layer docs that they use something like `data` for the endpoint. So perhaps it's still worth bikeshedding. But no need to before this PR gets merged.

tesseract-server/src/handlers/logic_layer/aggregate.rs
```diff
@@ -0,0 +1,222 @@
...
+                )
+            );
+        },
+    };
+
+    info!("format: {:?}", format);
+
+    let query = req.query_string();
+
+    let agg_query_res = serde_urlencoded::from_str::<Vec<(String, String)>>(query);
```
hwchen: I still strongly prefer the style:
```
struct LogicLayerQueryOpt {
  cube: String,
  dimensions: String,
  measures: String,
  #[serde(flatten)]
  cuts: HashMap<String, String>,
}
```
because
1) it provides a clearer view for maintainers to know what's in a query (I think data structures show this more clearly)
2) it provides one point of error handling, at deserialization, isntead of checking for each key

However, I leave this to you as I don't think there's anything wrong with using `HashMap` for all query params either.

hwchen: Oh, I think I also mentioned above, is there a reason you can't use `serde_qs`?

tesseract-server/src/schema_config.rs
MarcioPorto: Do we already have docs on writing schemas? I feel like that's where it should go, right?

tesseract-server/src/logic_layer/cache.rs
MarcioPorto: I need to refactor this file anyway. I'll wait until then.

hwchen:
One general comment: the boxed error helper is probably helpful across tesseract; it might be good to put in either the [server] or [core] crates.

hwchen:
Thanks for all your hard work on this! I know there was a lot of back-and-forth, but I'm feeling very comfortable with the direction this is going. Looking forward to more code :)

PR #89, opened 2019-03-07
Serde XML updates

MarcioPorto:
- Updates `serde-xml-rs` to version 0.3.1
- Adds support for annotations

### DIFF
--- a/tesseract-core/src/schema/xml.rs
+++ b/tesseract-core/src/schema/xml.rs
@@ -23,6 +23,8 @@ pub struct SchemaConfigXML {
     pub shared_dimensions: Option<Vec<SharedDimensionConfigXML>>,
     #[serde(rename(deserialize="Cube"))]
     pub cubes: Vec<CubeConfigXML>,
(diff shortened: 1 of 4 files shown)


tesseract-core/src/schema/xml.rs
```diff
@@ -97,12 +113,15 @@ pub struct TableConfigXML {
...
     pub column: String,
+    #[serde(rename(deserialize="Annotation"))]
+    pub annotations: Option<Vec<AnnotationConfigXML>>,
 }
 
 #[derive(Debug, Clone, PartialEq, Deserialize, Serialize)]
 pub struct AnnotationConfigXML {
     pub name: String,
-    pub label: String,
+    #[serde(rename(deserialize="$value"))]
```
hwchen: Curious, what is `$value`? is this specific to serde-xml?

hwchen:
Thanks!

Can you give an example of what the annotation syntax will look like in xml?

PR #90, opened 2019-03-08
[all] add contributors and license

hwchen:


MarcioPorto:
@hwchen Looks good on my end! 👍

jspeis:
sounds good 👍

LICENSE-MIT
```diff
@@ -0,0 +1,25 @@
+Copyright (c) 2016 keyring Developers
```
jspeis: one tweak though the license file should probably replace "keyring" with "tesseract"?

jspeis: and year to 2019

hwchen: Thanks for the catch :)

MarcioPorto: I don't know if you can test it before you merge this, but I think if you name the file as something like `LICENSE.md` GitHub picks it up and shows in the project navigation on the main page.

PR #95, opened 2019-03-15
Logic layer caption support

MarcioPorto:


hwchen:
I believe that the schema does not need to be changed; any property of a level can be a caption for that level.

hwchen:
thanks!

Issue #82, opened 2019-03-06
implement connection pooling for postgres


jspeis:
looks like bb8 recently updated to support the new tokio-postgres, would be nice to integrate with that if possible

jspeis:
done!

PR #103, opened 2019-03-20
Connection pooling support for postgres

jspeis:
Adds support for connection pooling in postgres using bb8 for #82


hwchen:
As a note, I'll probably clean up the DI with tokio-postgres afterwards, so that initializing the Pool can just return a future in the tesseract-postgres crate.

jspeis:
Sounds good, thanks!

PR #112, opened 2019-03-26
Feature/custom errors debug mode

hwchen:
- add custom error, which enables outputting db error as 500
- add debug mode switch, for controlling how errors surfaced


hwchen:
in `main.rs`, the logic checks out for dealing with debug from env var or cli.

- cli debug should always override if true
- if cli debug is false (meaning flag was not set), then should check if env var was set (result is Ok), and if it is set use that

Let me know if you think there's a better way to express this logic (or if you have other questions). Otherwise I'll probably just merge this tomorrow morning.

pinging @jspeis @MarcioPorto

jspeis:
Looks ok to me!

PR #116, opened 2019-04-03
Inline tables support

MarcioPorto:


tesseract-clickhouse/src/sql/primary_agg.rs
```diff
@@ -40,7 +41,13 @@ pub fn primary_agg(
     // So just swap the primary key DimSubquery to the head
 
     let mut ext_drills: Vec<_> = drills.iter()
-        .filter(|d| d.table.name != table.name)
+        .filter(|d| {
```
hwchen: hmmm... I feel like all inline tables should be external drills. If the inline table alias is the same as the fact table name, I would expect either
1) inline table should be further aliase (perhaps append `inline_table` or something)
2) there's an error during reading the schema, to make sure the names are not the same.

hwchen:
One overall comment: do you have an example of the config?

MarcioPorto:
```xml
                <InlineTable alias="tesseract_webshop_categories">
                    <ColumnDef name="category_name" key_type="text" />
                    <ColumnDef name="category_id" key_type="nontext" key_column_type="Int32" />

                    <Row>
                        <Value column="category_name">Inline Category: Books </Value>
                        <Value column="category_id">1</Value>
                    </Row>
                    <Row>
                        <Value column="category_name">Inline Category: Sports</Value>
                        <Value column="category_id">2</Value>
                    </Row>
                    <Row>
                        <Value column="category_name">Inline Category: Various</Value>
                        <Value column="category_id">3</Value>
                    </Row>
                    <Row>
                        <Value column="category_name">Inline Category: Videos</Value>
                        <Value column="category_id">4</Value>
                    </Row>
                </InlineTable>
```

tesseract-clickhouse/src/sql/primary_agg.rs
```diff
@@ -40,7 +41,13 @@ pub fn primary_agg(
...
+            if let Some(it) = &d.inline_table {
```
hwchen: Just want to note here, that you can use `if it.is_some()`, but no change necessary.

tesseract-server/src/schema_config.rs
```diff
@@ -19,13 +19,27 @@ pub fn read_schema(schema_path: &String) -> Result<Schema, Error> {
...
     }
 
-    // Check each cube for unique level and property names
     for cube in schema.cubes.clone() {
         let mut levels = HashSet::new();
         let mut properties = HashSet::new();
 
         for dimension in cube.dimensions.clone() {
             for hierarchy in dimension.hierarchies.clone() {
+                let has_table = match hierarchy.table {
```
hwchen: These can also be `.is_some()`

hwchen:
Looks good!

PR #119, opened 2019-04-11
Database name specification for ClickHouse

MarcioPorto:


hwchen:
`from_url` calls `Options::from_str`, which I would have guessed would automatically handle the parsing of the database, as well as the address. If this is true, I think that we can just deprecate the old `from_address`

tesseract-clickhouse/src/lib.rs
```diff
@@ -28,7 +30,27 @@ impl Clickhouse {
...
+
+        let pool = Pool::new(options);
+
+        Ok(Clickhouse {
+            pool,
+        })
+    }
+
+    pub fn from_url(url: &str) -> Result<Self, Error> {
+        let options = match Options::from_str(
```
hwchen: I think that you can use `parse` and `?` here.

`Options::from_str` is usually written as `let x: Options = "".parse()`, or with turbofish `let x = "".parse::<Options>()`.

the match statement is basically the old `try!` which is now `?`. failure's `Error` is a type which most other errors can be converted into (with early return) using `?`.

So the end product might be something like `let options = format!("tcp://{}", url).parse::<Options>()?`

PR #122, opened 2019-04-18
allow non-unique level names

hwchen:
Previously, for the sake of the logic layer, we had allowed only unique
level names in the schema.

Now, because we want to be able to have non-unique levels in the raw
schema, we just do a runtime check which changes what routes are loaded.

If unique levels, then regular logic layer is loaded.
If not unique levels, then a route which always returns 500 is loaded
for logic layer.

In the future, the LL will have aliases in the config to allow a
combination of LL and raw schema to have unique levels.


hwchen:
@MarcioPorto can you take a look?

MarcioPorto:
@hwchen Looks good!

PR #126, opened 2019-04-24
[logic layer] Adds name substitutions to logic layer config

MarcioPorto:
Example configuration, assuming there is a `[Category].[Category].[Category]` level and a `[Category].[Category].[Category].[Name]` property:

```json
    "aliases": {
        "cubes": [
            {
                "name": "Sales",
                "alternatives": ["sales", "sale"],
                "levels": [
                    {
                        "current_name": "Category 2.Category 2.Category",
                        "unique_name": "Category 2"
                    }
                ],
                "properties": [
                    {
                        "current_name": "Category 2.Category 2.Category.Name",
                        "unique_name": "Name 2"
                    }
                ]
            }
        ]
    }
```

### DIFF
--- a/tesseract-server/src/logic_layer/config.rs
+++ b/tesseract-server/src/logic_layer/config.rs
@@ -17,12 +17,20 @@ pub struct LogicLayerConfig {
 #[derive(Debug, Clone, Deserialize)]
 pub struct AliasConfig {
     pub cubes: Option<Vec<CubeAliasConfig>>,
+    pub shared_dimensions: Option<Vec<SharedDimensionAliasConfig>>
 }
 
 #[derive(Debug, Clone, Deserialize)]
 pub struct CubeAliasConfig {
     pub name: String,
     pub alternatives: Vec<String>,
+    pub levels: Option<Vec<LevelPropertyConfig>>,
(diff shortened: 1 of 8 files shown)


tesseract-server/src/logic_layer/config.rs
```diff
@@ -138,4 +158,82 @@ impl LogicLayerConfig {
...
+        for cube in &schema.cubes {
+            let mut levels = HashSet::new();
+            let mut properties = HashSet::new();
+
+            for dimension in &cube.dimensions {
+                for hierarchy in &dimension.hierarchies {
+
+                    // Check each cube for unique level and property names
+                    for level in &hierarchy.levels {
+                        let level_name = format!(
```
hwchen: Looks like you're matching on the string of the fully qualified name (I think? let me know if this isn't true).

It would be more ergonomic for the user if the `level` field they filled in in the config got parsed to a `LevelName` and then you could check that `LevelName` are equal.

The benefits would be giving an error on attempting to parse an ill-formed name, and allowing the user to enter a non-qualified name (just dim and level).

hwchen: Oh, this would also allow not using brackets in the name.

PR #130, opened 2019-05-06
Rate calculation

MarcioPorto:
This first version only allows 1 measure defined when a rate param is received. That measure is required to have either a sum or count aggregator and is used to calculate the rate.

Other than that, the raw Tesseract syntax is `rate=Dim.Hier.Level.val_1` and the logic layer syntax is `rate=Level.val_1`.

Multiple values can be passed in separated by commas: `rate=Level.val_1,val_2`.

For now, at least, we'll allow drilldowns and cuts on the level used for the rate calculation. Drilldowns will always result in a rate of 1. Cuts allow for rate calculation in a subset of the level universe (for example, calculating the rate of "Non-fiction" sales inside a "Books" named set.

### DIFF
--- a/tesseract-clickhouse/src/sql/primary_agg.rs
+++ b/tesseract-clickhouse/src/sql/primary_agg.rs
@@ -167,9 +167,12 @@ pub fn primary_agg(
         fact_sql.push_str(&format!(" where {}", cut_clause));
 
         if let Some(r) = rate {
-            // TODO: Allow for multiple members
             rate_fact_sql.push_str(
-                &format!(" where {} = {} and {}", r.column.clone(), r.members.clone(), cut_clause)
+                &format!(" where {} in ({}) and {}",
+                     r.column.clone(),
+                     join(r.members.clone(), ", "),
+                     cut_clause
+                )
             );
         }
     }
--- a/tesseract-clickhouse/src/sql/primary_agg.rs
+++ b/tesseract-clickhouse/src/sql/primary_agg.rs
@@ -132,18 +132,6 @@ pub fn primary_agg(
     let mut fact_sql = format!("select {}", all_fact_dim_cols);
     fact_sql.push_str(&format!(", {} from {}", mea_cols, table.name));
 
-    let rate_aggregator = match meas[0].aggregator {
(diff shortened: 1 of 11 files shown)


tesseract-clickhouse/src/sql/primary_agg.rs
```diff
@@ -126,10 +130,19 @@ pub fn primary_agg(
...
+        Aggregator::Count => "count".to_string(),
+        _ => "sum".to_string()
+    };
 
-    fact_sql.push_str(
-        &format!(", {} from {}", mea_cols, table.name)
-    );
+    let mut rate_fact_sql = "".to_string();
+
+    if let Some(_r) = rate {
```
hwchen: small comment: this could be `if rate.is_some()`

tesseract-clickhouse/src/sql/primary_agg.rs
```diff
@@ -207,6 +229,17 @@ pub fn primary_agg(
             sub_queries,
             dim_subquery.foreign_key
         );
+
+        // Wrap with rate subquery if there is a rate calculation
```
hwchen: If rate is going to wrap the primary aggregation, would it be possible to spin it into its own module? I had wondered about this since you were also having to modify rca.

For example, growth is a separate module, which wraps primary agg or rca.

tesseract-clickhouse/src/sql/rate.rs
```diff
@@ -0,0 +1,100 @@
...
+) -> (String, String)
+{
+    // Add a drilldown on the level we are getting the rate for
+    let mut new_drills: Vec<DrilldownSql> = vec![];
+    let mut found_rate_drill = false;
+
+    for drill in drills {
+        if drill == &rate.drilldown_sql {
+//            found_rate_drill = true;
+
```
hwchen: Let's remove these println too

hwchen:
@MarcioPorto thanks for the restructuring. Looks good and clean to me, I'll merge once the small fixes I noted are done.

Also, can you add a note somewhere outside of this pr that rate currently only allows one measure?

PR #136, opened 2019-05-09
[core][clickhouse] simple grouped median

hwchen:


hwchen:
Not feeling great, so just putting the feature up without testing yet.

schema syntax:
```
 "measures": [
  { "name": "Test Grouped Median", "column": "test", "aggregator": { "grouped_median": {"sub_aggregator": "sum"} }
 }
```

PR #139, opened 2019-05-22
Feature/default hierarchy

hwchen:


tesseract-core/src/lib.rs
```diff
@@ -202,9 +283,14 @@ impl Schema {
         let table = self.cube_table(&cube)
             .ok_or(format_err!("No table found for cube {}", cube))?;
 
-        let cut_cols = self.cube_cut_cols(&cube, &query.cuts)
+        let mut cut_cols = self.cube_cut_cols(&cube, &query.cuts)
+            .map_err(|err| format_err!("Error getting cut cols: {}", err))?;
+
+        let default_hierarchy_cut_cols = self.cube_cut_cols(&cube, &default_hierarchy_cuts_query)
```
MarcioPorto: A small thing, but maybe mention in the error message that this failed for the default hierarchy cut cols, instead of the regular cut cols.

tesseract-core/src/lib.rs
```diff
@@ -140,6 +178,8 @@ impl Schema {
         query: &Query,
         ) -> Result<(QueryIr, Vec<String>), Error>
     {
+        // TODO check that cuts have members
```
MarcioPorto: Saw you left this `TODO` here. Sounds like check you added in line 147 of `tesseract-clickhouse/src/sql/primary_agg.rs` so thought maybe you meant to remove this line.

hwchen: I actually meant that I should check that cuts have members at the beginning of `sql_query`, (or maybe on cut parsing?), to make clear that blank members will trigger default hierarchy behavior in sql generation. I'll improve this comment.

tesseract-core/src/lib.rs
hwchen: good point

PR #146, opened 2019-06-05
adds cube and dimension name to default hierarchy error message

jspeis:
Currently, when a tesseract user has a dimension with multiple hierarchies and doesn't specify a default, the error doesn't say where the issue is happening. So this PR adds a little extra info to make it easier to track down.

(Not crazy about the extra `cube_name = &cube.name` but was the easiest interim fix for working around double borrowing issue inside closure -- so feel free to suggest alternatives)


hwchen:
I think your borrowing fix is fine, I don't have a better alternative at the moment. I do agree it looks a little weird, but shouldn't be an issue.

PR #159, opened 2019-06-28
[core] refactor default member bugfix to error on incorrect default m…

hwchen:
…ember, but not when no default member

Let me know if the logic works. But I think what you want is that the error is propagated upwards only when there's an error parsing a `Cut` from the default member value.

if there's no default member value, just skip.

Previously, it was returning an error when there was a missing default member.


jspeis:
yup this looks like it works.

PR #162, opened 2019-07-29
Logic Layer cut operations

### LINKED ISSUE #163: [logiclayer] First pass at supporting cut operations
- [x] parents
- [x] children
- [x] neighbors

### LINKED ISSUE #164: [logiclayer] Store parents, children and neighbors IDs in cache

MarcioPorto:
closes #163 
closes #164


tesseract-clickhouse/src/lib.rs
```diff
@@ -74,15 +74,15 @@ impl Backend for Clickhouse {
         Box::new(fut_stream)
     }
 
+    // https://users.rust-lang.org/t/solved-is-it-possible-to-clone-a-boxed-trait-object/1714/4
```
hwchen: maybe a silly question :) but I'm not sure why this method was moved?

hwchen:
A general comment: I think we're going to try to move towards using logging macros vs `println`. I'd expect everything that's just an error from the user (e.g. duplicate drilldowns), or normal logging, to be `info!` level, with program failures being either `warn` or `error` (like unable to connect to db).

Some more verbose stuff can also be put in `debug`; tesseract generally runs at `info` so `debug` won't be logged.

@jspeis any other thoughts?

tesseract-core/src/schema.rs
```diff
@@ -274,6 +288,32 @@ impl Cube {
         }
         None
     }
+
+    pub fn get_child_level(&self, level_name: &LevelName) -> Result<Option<Level>, Error> {
+        let hierarchy =  match self.get_hierarchy(level_name) {
+            Some(h) => h,
+            None => {
+                return Err(format_err!("Could not find parent hierarchy for level: {}", level_name.level))
```
hwchen: This is a little shorter using `self.get_hierarchy().ok_or_else(|| format_err!())?`

tesseract-core/src/lib.rs
```diff
@@ -148,12 +148,20 @@ impl Schema {
                     // Check each cube for unique level and property names
                     for level in &hierarchy.levels {
                         if !levels.insert(&level.name) {
+                            println!(
```
hwchen: see general comment about logs and println

hwchen:
Nice docs!

tesseract-server/src/logic_layer/util.rs
```diff
@@ -0,0 +1,133 @@
...
+        future::result(
+            Ok(HttpResponse::NotFound().json(message))
+        )
+    )
+}
+
+
+/// DataFrame columns can come in many different types. This function converts
+/// all data to a common type (String).
+pub fn stringify_column_data(col: &Column) -> Vec<String> {
```
hwchen: Would it be good to move this to the Dataframe module?

tesseract-server/src/logic_layer/util.rs
```diff
@@ -0,0 +1,133 @@
...
+    FutureResponse,
+    HttpResponse,
+};
+use futures::future::{self};
+
+use tesseract_core::{ColumnData, Column};
+
+
+/// Helper method to return errors (FutureResponse<HttpResponse>).
+pub fn boxed_error(message: String) -> FutureResponse<HttpResponse> {
```
hwchen: can this be moved to a more general `util` module?

tesseract-server/src/handlers/logic_layer/aggregate.rs
```diff
@@ -71,124 +133,188 @@ pub fn logic_layer_aggregation(
...
+            match llc.sub_cube_name(agg_query.cube.clone()) {
+                Ok(cn) => cn,
+                Err(_) => agg_query.cube.clone()
+            }
+        },
+        None => agg_query.cube.clone()
+    };
 
-            let cube_cache = req.state().cache.read().unwrap().find_cube_info(&cube_name).clone();
+    let cube = match schema.get_cube_by_name(&cube_name) {
```
hwchen: I forget if `get_cube_by_name` returns a reference to a cube. But if it did, would it be useable for your purposes in the rest of this method?

tesseract-server/src/handlers/logic_layer/aggregate.rs
```diff
@@ -71,124 +133,188 @@ pub fn logic_layer_aggregation(
...
+        let (query_ir, headers) = match query_ir_headers {
+            Ok(x) => x,
+            Err(err) => return boxed_error(err.to_string())
+        };
 
-                let time = match Time::from_key_value(tc[0].clone(), tc[1].clone()) {
-                    Ok(time) => time,
-                    Err(err) => return boxed_error(err.to_string())
-                };
+        info!("Query IR: {:?}", query_ir);
```
hwchen: do you think this could be debug, or should it be kept as info?

tesseract-server/src/handlers/logic_layer/aggregate.rs
```diff
@@ -71,124 +133,188 @@ pub fn logic_layer_aggregation(
...
-    // Turn AggregateQueryOpt into TsQuery
-    let ts_query: Result<TsQuery, _> = agg_query.try_into();
-    let ts_query = match ts_query {
-        Ok(q) => q,
-        Err(err) => return boxed_error(err.to_string())
-    };
+    info!("Headers: {:?}", final_headers);
+
+    // Joins all the futures for each TsQuery
+    let futures: JoinAll<Vec<Box<Future<Item=DataFrame, Error=Error>>>> = join_all(sql_strings
```
hwchen: small nitpick: because `futures` is also the name of the module, this might cause a naming conflict in the future. I'd prefer `futs`, but let me know what you think.

tesseract-server/src/handlers/logic_layer/aggregate.rs
```diff
@@ -71,124 +133,188 @@ pub fn logic_layer_aggregation(
...
+            let mut final_columns: Vec<Column> = vec![];
+            let num_cols = dfs[0].columns.len();
+
+            for col_i in 0..num_cols {
+                let mut same_type = true;
+                let mut curr_type: i32 = -1;
+
+                for df in &dfs {
+                    let c = &df.columns[col_i];
+                    let i = match c.column_data {
```
hwchen: can you walk me through this section in person? First glance, changing an enum to ints (kind of like C or Go enums) seems strange, but I might not be understanding your needs here.

Also, if it is important, then perhaps some of this can be put into the Dataframe module.

tesseract-server/src/handlers/logic_layer/aggregate.rs
```diff
@@ -206,3 +332,682 @@ pub fn logic_layer_aggregation(
         })
         .responder()
 }
+
+
+/// Generates a series of Tesseract queries from a single LogicLayerQueryOpt.
+/// This function contains the bulk of the logic layer logic.
+pub fn generate_ts_queries(
```
hwchen: Since this diff is so big, just want to check here if you made big changes to this fn. I feel like I remember some of this code from before.

tesseract-server/src/handlers/logic_layer/aggregate.rs
```diff
@@ -206,3 +332,682 @@ pub fn logic_layer_aggregation(
...
+    Ok(agg_query_opt_cuts)
+}
+
+
+/// Implements logic to resolve logic layer cuts (including those with operations)
+/// into a HashMap separating cuts for each dimension. Doing so helps generate all
+/// the possible cut combinations in the next step.
+/// This method also returns a HashMap of header name substitutions that will help
+/// with the naming of the final column names in the response.
+pub fn resolve_cuts(
```
hwchen: Also just checking if there's any big changes here since it's a big diff.

tesseract-server/src/handlers/logic_layer/aggregate.rs
```diff
@@ -206,3 +332,682 @@ pub fn logic_layer_aggregation(
...
+
+    Ok((queries, header_map))
+
+}
+
+
+/// Given a vector containing a partial Cartesian product, and a list of items,
+/// return a vector adding the list of items to the partial Cartesian product.
+/// From: https://gist.github.com/kylewlacy/115965b40e02a3325558
+pub fn partial_cartesian<T: Clone>(a: Vec<Vec<T>>, b: Vec<T>) -> Vec<Vec<T>> {
```
hwchen: can the cartesian fns get put into the util module? Also, just wanted to double check that this functionality can't be provided by itertools: https://docs.rs/itertools/0.8.0/itertools/trait.Itertools.html#method.cartesian_product

tesseract-server/src/logic_layer/cache.rs
```diff
@@ -5,12 +5,101 @@ use log::info;
```
hwchen: Everything here appears to be good, but it is a pretty big diff :) I think that if there's errors, they will probably need to be shaken out (are there any places to test?), and otherwise the structure feels alright.

tesseract-clickhouse/src/lib.rs
MarcioPorto: I kept getting a warning that the order of the methods in here was different from the trait definition in https://github.com/hwchen/tesseract/blob/master/tesseract-core/src/backend.rs. Just wanted to get rid of it 😄

tesseract-server/src/handlers/logic_layer/aggregate.rs
MarcioPorto: Not to this part of the function, just some small cleanup.

tesseract-server/src/handlers/logic_layer/aggregate.rs
MarcioPorto: I moved the core of the cut resolution logic out of the `generate_ts_queries` function into this one because it was getting too big. So here is where I am doing a lot of the logic for dealing with cut operations.

tesseract-server/src/logic_layer/cache.rs
MarcioPorto: Can you elaborate on your comment about shaking out errors?

tesseract-server/src/handlers/logic_layer/aggregate.rs
MarcioPorto: It does return a reference. Are you just saying I should remove the clone?

tesseract-server/src/handlers/logic_layer/aggregate.rs
MarcioPorto: I don't think these are quite equivalent. The `itertools` method generates pairs between two iterators. We're doing something a little different here.

tesseract-server/src/handlers/logic_layer/aggregate.rs
hwchen: is it possible to remove the clone? I guess if you're just cloning a ref it's not as big a deal. But it did give me pause.

tesseract-server/src/logic_layer/cache.rs
hwchen: oh, it's just about logic errors that might be hard to catch until the user makes a particular request. Like duplicate drilldowns, etc.

tesseract-server/src/handlers/logic_layer/aggregate.rs
MarcioPorto: I already removed the clone here: https://github.com/tesseract-olap/tesseract/pull/162/commits/c9751375aca7e20d494c6e3a877c136d601d82fb#diff-a1b1aab7cc043def804944a357fb1fedL154

tesseract-core/src/dataframe.rs
```diff
@@ -65,6 +65,119 @@ impl Column {
...
         &mut self.column_data
     }
+
+    /// DataFrame columns can come in many different types. This function converts
+    /// all data to a common type (String).
+    pub fn stringify_column_data(&self) -> Vec<String> {
+        return match &self.column_data {
+            ColumnData::Int8(v) => {
+                let mut t: Vec<i8> = v.iter().map(|&e| e.clone() as i8).collect();
+                t.sort();
```
hwchen: leaving a note here for tracking, after discussing with @MarcioPorto in person; sort on dataframe cols will be pulled into a separate method.

hwchen:
Looks good! Thanks!

PR #174, opened 2019-08-08
Geoservice support

### LINKED ISSUE #173: [logic layer] Add Geoservice support for cut resolutions

MarcioPorto:
closes #173

### DIFF
--- a/tesseract-server/src/handlers/logic_layer/aggregate.rs
+++ b/tesseract-server/src/handlers/logic_layer/aggregate.rs
@@ -1,4 +1,5 @@
 use std::collections::HashMap;
+use std::str;
 
 use actix_web::{
     client,
@@ -923,7 +924,7 @@ pub fn resolve_cuts(
                         DimensionType::Geo => {
                             match geoservice_url {
                                 Some(geoservice_url) => {
-                                    let neighbors_ids = perform_geoservice_query(geoservice_url, &elements[0]);
+                                    let neighbors_ids = query_geoservice(geoservice_url, &elements[0]);
 
                                     println!(" ");
                                     println!(" ");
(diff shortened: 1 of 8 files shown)


tesseract-server/src/handlers/logic_layer/aggregate.rs
```diff
@@ -908,23 +915,44 @@ pub fn resolve_cuts(
...
-                        // TODO: Add a better way to identify geography levels in the schema
-                        // TODO: Wait for geoservice API to be ready
-                        return Err(format_err!("Geoservice neighbors not currently supported."));
-                    } else {
-                        let level_cache = match cube_cache.level_caches.get(&level_name.level) {
-                            Some(level_cache) => level_cache,
-                            None => return Err(format_err!("Could not find cached entries for {}.", level_name.level))
-                        };
+                    // Find dimension for the level name
+                    let dimension = match cube.get_dimension(&level_name) {
```
hwchen: this could be a place for `map_err`, I think. But it's up to you.

tesseract-server/src/handlers/logic_layer/geoservice.rs
```diff
@@ -0,0 +1,82 @@
...
+
+/// Queries geoservice for geo cuts resolution.
+pub fn query_geoservice(
+    geoservice_url: &str,
+    geoservice_query: &GeoserviceQuery,
+    geo_id: &str
+) -> Result<Vec<GeoServiceResponseJson>, Error> {
+    let mut query_url: String;
+
+    if &geoservice_url[geoservice_url.len()-1..] == "/" {
```
hwchen: you can do `let query_url =` and then the whole match block, instead of initializing then assigning.

hwchen: Also, I think that you should use the Url crate (which is probably reexported through actix) in order to handle the url robustly. Let me know what you think.

tesseract-server/src/handlers/logic_layer/aggregate.rs
```diff
@@ -908,23 +915,44 @@ pub fn resolve_cuts(
...
-                            Some(neighbors_ids) => neighbors_ids.clone(),
-                            None => continue
-                        };
+                    match dimension.dim_type {
+                        DimensionType::Geo => {
+                            match geoservice_url {
+                                Some(geoservice_url) => {
+                                    let mut neighbors_ids: Vec<String> = vec![];
+
+                                    let geoservice_response = query_geoservice(geoservice_url, &GeoserviceQuery::Neighbors, &elements[0])?;
```
hwchen: Currently, you're explicitly waiting for the geoservice. This means that a slow geoservice call would block the thread. However, I could see that doing this if/then logic using combinators could be pretty involved, so I'm willing to let it through for now. But this should be a high priority refactor when async/await comes through.

hwchen: If we migrate to actix 1.0, this could also be put inside a `block` section

MarcioPorto: https://github.com/tesseract-olap/tesseract/issues/181

tesseract-server/src/handlers/logic_layer/geoservice.rs
```diff
@@ -31,23 +32,16 @@ pub fn query_geoservice(
     geoservice_query: &GeoserviceQuery,
     geo_id: &str
 ) -> Result<Vec<GeoServiceResponseJson>, Error> {
-    let mut query_url: String;
+    let base_url = Url::parse(geoservice_url.clone()).unwrap();
```
hwchen: Let's move the parsing to tesseract startup.
There, it should be easy to return a good error.
Also, parse takes `&str`, so I don't think you have to clone?

tesseract-server/src/handlers/logic_layer/geoservice.rs
```diff
@@ -31,23 +32,16 @@ pub fn query_geoservice(
...
-    } else {
-        match geoservice_query {
-            GeoserviceQuery::Neighbors => query_url = format!("{}/neighbors/{}", geoservice_url, geo_id),
-            GeoserviceQuery::Children => query_url = format!("{}/relations/children/{}", geoservice_url, geo_id),
-            GeoserviceQuery::Parents => query_url = format!("{}/relations/parents/{}", geoservice_url, geo_id),
-            _ => return Err(format_err!("This type of geoservice query is not yet supported"))
-        }
-    }
+    let join_str = match geoservice_query {
+        GeoserviceQuery::Neighbors => format!("{}neighbors/{}", geoservice_url, geo_id),
```
hwchen: geoservice_url is being passed in to this str, won't that make it repeat because you also join later?

tesseract-server/src/handlers/logic_layer/geoservice.rs
```diff
@@ -31,23 +32,16 @@ pub fn query_geoservice(
...
+        GeoserviceQuery::Children => format!("{}relations/children/{}", geoservice_url, geo_id),
+        GeoserviceQuery::Parents => format!("{}relations/parents/{}", geoservice_url, geo_id),
+        _ => return Err(format_err!("This type of geoservice query is not yet supported"))
+    };
+
+    let query_url = base_url.join(&join_str).unwrap().as_str().to_string();
```
hwchen: does it work if you remove `to_string()`, but pass to client as `&query_url`?

tesseract-server/src/handlers/logic_layer/geoservice.rs
MarcioPorto: Good call! Somehow the join was able to fix that for me so I didn't even notice!

tesseract-server/src/handlers/logic_layer/geoservice.rs
MarcioPorto: Screen Shot 2019-08-14 at 11 53 27 AM

This is what I get when I remove the `to_string()`

hwchen:
Thanks! and thanks for your patience with all the requested changes.

MarcioPorto:
No problem, your suggestions were great! 😃

PR #178, opened 2019-08-13
[core] first pass on measuretype

hwchen:
@frabarz adding you on as a reviewer because I wanted to ask some questions. Mainly, to confirm whether you want the structure of `standard` and `error` to be the same; then I'll add a field like `units` or something.

### DIFF
--- a/tesseract-core/src/schema.rs
+++ b/tesseract-core/src/schema.rs
@@ -602,6 +602,7 @@ pub struct Measure{
     pub name: String,
     pub column: String,
     pub aggregator: Aggregator,
+    pub measure_type: MeasureType,
     pub annotations: Option<Vec<Annotation>>,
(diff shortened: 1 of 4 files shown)


hwchen:
nope, when it's standard, there can be an object, just like for error. So something like:
```
"measure_type": {
  "standard": {
    "units": "kilograms" // or null here, units field will always render
  }
}
```

as for overlap with aggregator, there's always some relation but it shouldn't be too coupled; measure_type is more about external representation, and aggregator is more an internal implementation detail. I guess it's kind of confusing, but we're asking the schema to do both ways.

hwchen:
I actually have to implement my last comment :) but that will be the final structure for now.

hwchen:
rebased on master.

PR #183, opened 2019-08-19
Members locale enpoint

### LINKED ISSUE #175: [logic layer] Add members endpoint with locale support

### LINKED ISSUE #182: [logic layer] Add caption set support for inline table rows
Inline tables currently can't support locale queries because their rows don't accept a `caption_set` parameter like a `Property` object does.

MarcioPorto:
closes #175 
closes #182

### DIFF
--- a/tesseract-core/src/lib.rs
+++ b/tesseract-core/src/lib.rs
@@ -205,7 +205,82 @@ impl Schema {
         Ok((sql, header))
     }
 
-    /// Convert user parameters into required default member cuts based on cube defintion.
+    pub fn members_locale_sql(
+        &self,
+        cube_name: &str,
+        level_name: &LevelName,
+        locale: &str
+    ) -> Result<(String, Vec<String>), Error> // Sql and then Header
+    {
+        // TODO: Add support for multiple locales
+
+        let cube = self.cubes.iter()
+            .find(|cube| &cube.name == &cube_name)
+            .ok_or(format_err!("Could not find cube"))?;
+
+        let dim = cube.dimensions.iter()
+            .find(|dim| dim.name == level_name.dimension)
+            .ok_or(format_err!("could not find dimension for level name"))?;
+        let hier = dim.hierarchies.iter()
+            .find(|hier| hier.name == level_name.hierarchy)
+            .ok_or(format_err!("could not find hierarchy for level name"))?;
+        let level = hier.levels.iter()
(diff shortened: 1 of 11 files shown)


tesseract-core/src/lib.rs
```diff
@@ -205,7 +205,94 @@ impl Schema {
...
+            .find(|lvl| lvl.name == level_name.level)
+            .ok_or(format_err!("could not find level for level name"))?;
+
+        let table = hier.table.clone().unwrap_or_else(|| cube.table.clone());
+
+        let key_column = level.key_column.clone();
+        let mut header = vec!["ID".into()];
+        let mut name_columns: Vec<String> = vec![];
+
+        // TODO: have a check that there can't be inline table and regular table at the same time.
```
hwchen: As as note, I think I have somewhere else in the code a TODO to make table an enum instead of having different fields for in-line and regular tables.

tesseract-server/src/handlers/logic_layer/metadata.rs
```diff
@@ -0,0 +1,198 @@
...
+
+    let members_query = match QS_NON_STRICT.deserialize_str::<MembersQueryOpt>(query) {
+        Ok(q) => q,
+        Err(err) => return boxed_error(err.to_string())
+    };
+
+    let mut cube_name = members_query.cube.clone();
+    let mut level_name: Option<LevelName> = None;
+
+    match &logic_layer_config {
```
hwchen: We should probably investigate some ways to reduce `Option` nesting hell. In the meantime, can we use `if let Some` so that at least we don't have the `None` side of the match?

tesseract-server/src/handlers/metadata.rs
```diff
@@ -90,6 +90,10 @@ pub fn do_members(
         },
     };
 
+    println!(" ");
```
hwchen: switch to log?

tesseract-core/src/lib.rs
```diff
@@ -205,7 +205,94 @@ impl Schema {
...
+                        None => ()
+                    }
+                }
+            }
+
+            format!("({})", inline.sql_string())
+        } else {
+            for locale in &locales {
+                match &level.properties {
+                    Some(properties) => {
```
hwchen: `if let Some` here would remove need for `None` match (there's another comment mentioning the same)

hwchen:
thanks!

PR #189, opened 2019-08-29
[core] metadata first pass unique level name

hwchen:


hwchen:
ugh, looks like I have rustfmt on save here, I'll want to change it back.

hwchen:
Doesn’t do the thing we need.

PR #195, opened 2019-09-15
[bug] WIP fix to deduplicate drilldowns

AnthonyHewins:
This is definitely not done, but if you want to check if this is anywhere in the expected ballpark, please go ahead


jspeis:
@hwchen what do you think so far?

The only thing I might suggest adding so far would be a test case for the `Ok` case. Nice work so far @AnthonyHewins !

tesseract-core/src/query.rs
```diff
@@ -490,3 +504,22 @@ impl FromStr for RateQuery {
...
 }
+
+#[cfg(test)]
+mod test {
+    use super::*;
+
+    #[test]
+    fn test_validate() {
+        let mut q = Query::new();
+        assert!(q.validate().is_ok());
```
AnthonyHewins: @jspeis is this what you mean? General question too, should I be putting multiple assertions in one function or should I do one assertion per function, which one is better practice in Rust?

jspeis: @AnthonyHewins, something like that -- yes is what I had in mind with two minor notes to the example above exception: to be a valid query a query should have at least one drilldown and at least one measure (cc @hwchen) so technically I would expect the default query which has a empty vec for both drilldowns and measures to be _invalid_. 

In terms of which is a better practice, I think it's more a matter of style and preference. Overall I think it's good to have a test function focus on one task, which tends to make it easier to give a function a descriptive name that others will be able to follow and understand. In this particular case, I could see it going either way because the logic is not particularly complicated.

hwchen: It is a little flexible, but I generally have more than one assertion in a test function. It really depends on the test.

For this series of tests, I might do:
- validate_missing_params (which could include assertions for missing drilldown, missing measure, and both)
- validate_rca
- etc.
Let me know if this clarifies things.

hwchen: And yes, a valid query should have both a drilldown and a measure.

hwchen:
Thanks again @AnthonyHewins for your work, really sorry I've been slow to respond, I'm a bit busy at the moment.

AnthonyHewins:
Thanks for the comments, I just updated my PR. Added logic to check that measures and drilldowns both exist. Also made the assertions more verbose to make sure it was picking up on the right error.

Apologize if I'm lacking a bit of knowledge in the OLAP domain...I learned about them very briefly about 3-4 years ago and it wasn't in-depth. I will put aside some time to get up to speed, it's a realm I'm going to encounter anyways at some point.

jspeis:
@AnthonyHewins this is looking very good! I will take a closer look tomorrow and test it out

jspeis:
Looks good to me so far. @hwchen in terms of hooking this up, were you thinking that this would get called in `do_aggregate`?

hwchen:
yup @jspeis that sounds good.

jspeis:
@AnthonyHewins would you feel comfortable hooking up your `validate` function in the `do_aggregate` functions? (both in the regular and streaming modes?)

We're happy to help if you have any questions!

AnthonyHewins:
Absolutely, this weekend is probably when I can get around to doing that. I may have some questions along the way but I'll just post them after I make a commit

jspeis:
Great! And feel free to let us know if you have any questions

AnthonyHewins:
Sorry it's taken some time, things got a little crazy the past two weeks, but I have gotten around to implementing part of it. I want to spend some time with actix so I know how to properly send the right error with `FutureResponse` before I push up though, once I get a chance to reach that part in the tutorial I will push up and hopefully this drawn-out PR will come to a close lol

PR #196, opened 2019-09-16
Bug for default locales

jspeis:
Here is a proposed fix for handling captioning of labels with default languages

### DIFF
--- a/README.md
+++ b/README.md
@@ -50,7 +50,7 @@ Note that a systemd `.service` is also installed. You will probably need to modi
 
 ### Environment Variables
 - `TESSERACT_DATABASE_URL`: required, is the address of the database; make sure to include the user, password, and database name.
-- `TESSERACT_DEBUG`: boolean, `true` is a debug mode for data/calculations. As of now, the only feature is the `rca` calculation will show `a b c d` sub-aggregations for the calculation`(a/b)/(c/d)`.
+- `TESSERACT_DEBUG`: boolean, `true` is a flag to enable more verbose logging output to help the debugging process while testing.
 - `TESSERACT_FLUSH_SECRET`: optional, but required for flush; is the secret key for the flush endpoint.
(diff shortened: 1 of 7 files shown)


README.md
```diff
@@ -50,11 +50,12 @@ Note that a systemd `.service` is also installed. You will probably need to modi
 
 ### Environment Variables
 - `TESSERACT_DATABASE_URL`: required, is the address of the database; make sure to include the user, password, and database name.
-- `TESSERACT_SCHEMA_FILEPATH`: required, should point to the location on disk for the tesseract schema file.
-- `TESSERACT_LOGIC_LAYER_CONFIG_FILEPATH`: optional, should point to the location on path for the logic layer configuration.
+- `TESSERACT_DEBUG`: boolean, `true` is a debug mode for data/calculations. As of now, the only feature is the `rca` calculation will show `a b c d` sub-aggregations for the calculation`(a/b)/(c/d)`.
```
MarcioPorto: I think while we're at it we should update this text. I believe now setting this to true also returns mote detailed error messages in the response when queries fail, so it goes beyond just RCA.

jspeis:
@hwchen let me know if this looks good and then we can merge

tesseract-core/src/schema.rs
```diff
@@ -138,10 +139,13 @@ impl From<SchemaConfigJson> for Schema {
                     .collect()
             });
 
+        let default_locale_str: &str = "en";
```
hwchen: can we make this a `const DEFAULT_LOCALE_STR` and put it at the top of the schema? (that's my preference, anyways, let me know if you prefer this way of handling hardcoded stuff)

jspeis: @hwchen agreed! I've just updated this

hwchen:
Thanks!

PR #197, opened 2019-09-23
[clickhouse] update driver to 0.1.14

hwchen:


jspeis:
This looks good. Though I did just realize that my default_locale PR broke one of the tests so I'll fix that

PR #208, opened 2019-09-26
Feature/simple moe

hwchen:
- add simple moe agg, with basic formula of `sqrt(sum(power(moe / 1.645,
  2))) * 1.645`
- update previous moe to be called replicate weight moe.
- add critical value variable to replicate weight moe.
- add critical value variable to weighted avg moe.

### DIFF
--- a/tesseract-clickhouse/src/sql/aggregator.rs
+++ b/tesseract-clickhouse/src/sql/aggregator.rs
@@ -62,7 +62,8 @@ pub fn agg_sql_string_pass_1(col: &str, aggregator: &Aggregator, mea_idx: usize)
             )
         },
         Aggregator::Moe { .. }=> {
-            format!("sum(power( m{0} / 1.645, 2)) as m{0}_moe_sub_agg",
+            format!("sum(power( {} / 1.645, 2)) as m{}_moe_sub_agg",
+                col,
                 mea_idx,
             )
         },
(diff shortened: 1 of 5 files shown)


hwchen:
closes #209

tesseract-clickhouse/src/sql/aggregator.rs
```diff
@@ -61,6 +61,12 @@ pub fn agg_sql_string_pass_1(col: &str, aggregator: &Aggregator, mea_idx: usize)
                 join(secondaries, ", "),
             )
         },
+        Aggregator::Moe { .. }=> {
+            format!("sum(power( {} / 1.645, 2)) as m{}_moe_sub_agg",
```
MarcioPorto: Maybe I am missing something, but should this `1.645` be substituted by the critical value?

tesseract-clickhouse/src/sql/aggregator.rs
```diff
@@ -111,7 +117,7 @@ pub fn agg_sql_string_select_mea(aggregator: &Aggregator, mea_idx: usize) -> Str
                 mea_idx,
             )
         },
-        Aggregator::Moe { secondary_columns, .. }=> {
+        Aggregator::ReplicateWeightMoe { secondary_columns, .. }=> {
```
jspeis: should we add a space between the `}` and the  `=>`?

tesseract-clickhouse/src/sql/aggregator.rs
hwchen: ah, yes, this one should be. thanks!

hwchen:
Will document

hwchen:
Added the beginnings of a docs folder. And there's examples for all the aggregators finally.

hwchen:
@jspeis @MarcioPorto just checking if there's anything else before I merge. I added some docs for the aggregators. (and the beginnings of a docs folder, which could in the future be used for a static site generator)

MarcioPorto:
Looks good to me!

hwchen:
closes #209

PR #215, opened 2019-10-08
Feature/member cache

### LINKED ISSUE #207: cache members for cut validation
To prevent sql injection, validate cut members against cached member list. Can use current logic layer cache.

hwchen:
This is a basic pass at using a members cache to be able to validate cut member IDs before sending it on to sql.

This needs some work before being robust enough to work
- without any logic layer involvement
- for a /members endpoint

However, in its current form, it may be usable enough for our requirements.

```
TODO note that this is being used in core tesseract, but that
the cache is created using logic layer rules. This means that
at the moment of this implementation, this will work in core
tesseract but only if the core tesseract schema can also be a
logic layer schema (and satisfying all the rules about level,
property uniqueness, etc.) The next step is to figure out how
to architecture this cache so that it can be used even in
non-logic layer setups, but also be used for members endpoint
(which requires ID)
```

closes #207


tesseract-server/src/logic_layer/cache.rs
```diff
@@ -225,6 +225,19 @@ impl CubeCache {
             None => None
         }
     }
+
+    // TODO note that this is being used in core tesseract, but that the cache is created using
+    // logic layer rules. This means that at the moment of this implementation, this will work in
+    // core tesseract but only if the core tesseract schema can also be a logic layer schema (and
+    // satisfying all the rules about level, property uniqueness, etc.)
```
jspeis: overall this looks great! what will happen if a user doesn't satisfy the rules about level/property uniqueness at the moment, what kind of error would they get?

jspeis:
@MarcioPorto I've pushed some minor tweaks, when you have a moment, can you take a look to see if this looks good to go?

tesseract-server/src/logic_layer/cache.rs
```diff
@@ -425,8 +447,8 @@ pub fn populate_cache(
         })
     }
 
-    info!("Cache ready!");
-
+    let timing = time_start.elapsed();
```
MarcioPorto: This is great. Thanks for adding it.

tesseract-server/src/handlers/aggregate.rs
```diff
@@ -240,3 +265,21 @@ impl TryFrom<AggregateQueryOpt> for TsQuery {
         })
     }
 }
+
+use failure::{bail, format_err};
+use tesseract_core::names::Cut;
+use crate::logic_layer::CubeCache;
+
+fn validate_members(cuts: &[Cut], cube_cache: &CubeCache) -> Result<(), Error> {
```
MarcioPorto: Should we move this somewhere where to logic layer could use it as well?

jspeis: Agreed! What do you think about having it in `handlers/util.rs`? (see latest push)

MarcioPorto: That looks good to me!

MarcioPorto:
@jspeis Made a few changes to also include this check to the logic layer aggregate method.

tesseract-server/src/logic_layer/cache.rs
jspeis: @MarcioPorto we should just review this together

PR #223, opened 2019-10-30
Added new logic layer relations endpoint

Jayrmajithia:
Adds an endpoint geoadjacents which can help in identifying the neighbors/parents/children of the given id.


tesseract-server/src/app.rs
```diff
@@ -24,6 +24,10 @@ use crate::handlers::{
     metadata_all_handler,
     members_handler,
     members_default_handler,
+    logic_layer_geoadjacents_handler,
```
MarcioPorto: As an overall theme, I think we should rename this endpoint to something a little broader than `geoadjacents` since it can support non-geo levels.

I am leaning towards something like `/relations` since that's broad enough to include everything, but I am open any ideas you might have.

tesseract-server/src/app.rs
```diff
@@ -24,6 +24,10 @@ use crate::handlers::{
...
+    logic_layer_default_geoadjacents_handler,
```
MarcioPorto: `logic_layer_geoadjacents_default_handler` for consistency

tesseract-server/src/handlers/logic_layer/geoadjacents.rs
```diff
@@ -0,0 +1,351 @@
...
+pub struct LogicLayerQueryOpt {
+    pub cube: String,
+    #[serde(flatten)]
+    pub cuts: Option<HashMap<String, String>>,
+    parents: Option<bool>,
+    debug: Option<bool>,
+    exclude_default_members: Option<bool>,
+    locale: Option<String>,
+    sparse: Option<bool>,
+}
```
MarcioPorto: Feel free to remove `parents` and `sparse` since they're not being used. `locale` can also be removed as it does not make sense here. And I would remove `exclude_default_members` as well (we can add it back in case we need to support it).

Lastly, this query opt should have a more descriptive name.

tesseract-server/src/handlers/logic_layer/geoadjacents.rs
```diff
@@ -0,0 +1,351 @@
...
+
+    let cube_cache = match req.state().cache.read().unwrap().find_cube_info(&cube_name) {
+        Some(cube_cache) => cube_cache,
+        None => return Ok(HttpResponse::NotFound().json("Unable to access cube cache".to_string()))
+    };
+
+    let cuts_map = match agg_query.cuts{
+        Some(cut_map) => cut_map,
+        None => HashMap::new(),
+    };
```
MarcioPorto: Instead of creating an empty HashMap here, we should return an error if no cuts are provided.

tesseract-server/src/handlers/logic_layer/geoadjacents.rs
```diff
@@ -0,0 +1,351 @@
...
+            Ok(HttpResponse::Ok()
+                .set(content_type)
+                .body(res))
+        },
+        Err(err) => Ok(HttpResponse::NotFound().json(err.to_string())),
+    }
+}
+
+
+pub fn getdimensions(
```
MarcioPorto: Rename this to something like `get_relations`

tesseract-server/src/handlers/logic_layer/geoadjacents.rs
```diff
@@ -0,0 +1,351 @@
...
+    let property_map = &cube_cache.property_map;
+    let geoservice_url = &req.state().env_vars.geoservice_url;
+
+    let dimensions_map = match getdimensions(&cuts_map, &cube, &cube_cache, &level_map, &property_map, &geoservice_url) {
+        Ok(dm) => dm,
+        Err(err) => return Ok(HttpResponse::NotFound().json(err.to_string()))
+    };
+
+    if dimensions_map.len() == 0{
+        return Ok(HttpResponse::NotFound().json("Unable to genrate result".to_string()));
```
MarcioPorto: I was talking to Jimmy and he and I agreed that, in this case, the response should just be an empty dataframe. So I don't think we need this error here. Can you also check that we don't get an error when the response comes back empty?

tesseract-server/src/handlers/logic_layer/geoadjacents.rs
```diff
@@ -0,0 +1,351 @@
...
+        }
+        let element:Vec<String> = cut_values.split(":").map(|s| s.to_string()).collect();
+        let cut_key_values:Vec<String> = match element.get(0){
+            Some(ckv) => ckv.split(",").map(|s| s.to_string()).collect(),
+            None => continue,
+        };
+        let operations:Vec<String> = match element.get(1){
+            Some(op) => op.split(",").map(|s| s.to_string()).collect(),
+            None => continue,
+        };
```
MarcioPorto: For consistency and ease of reading, we should add spaces before variable types. Similarly, we've been using spaces before open curly braces:

`let operations:Vec<String> = match element.get(1){` becomes `let operations: Vec<String> = match element.get(1) {`

We should set up a linter for this type of stuff, but we've been doing it manually for now...

tesseract-server/src/handlers/logic_layer/geoadjacents.rs
```diff
@@ -0,0 +1,351 @@
...
+                            match children_map.get(&cut) {
+                                Some(children_ids) => children_ids.clone(),
+                                None => continue
+                            }
+                        },
+                        None => continue
+                    };
+
+                    for children_id in children_ids.iter(){
+                        dimensions_map.push([level_name.level.to_string(), cut.to_string(), "children".to_string(), children_id.to_string()].to_vec())
```
MarcioPorto: I don't think this will work well for the level names. We need to return their unique names according to the logic layer, not the name of the level in their level name object. These two will not always be the same. Let me know if you need any clarification.

tesseract-server/src/handlers/logic_layer/geoadjacents.rs
```diff
@@ -0,0 +1,351 @@
...
+                            Some(parent_map) => {
+                                match parent_map.get(&cut) {
+                                    Some(parent_id) => parent_id.clone(),
+                                    None => continue
+                                }
+                            },
+                            None => continue
+                        };
+
+                        dimensions_map.push([level_name.level.to_string(), cut.to_string(), "parent".to_string(), parent_id.to_string()].to_vec());
```
MarcioPorto: Same as above.

tesseract-server/src/handlers/logic_layer/geoadjacents.rs
```diff
@@ -0,0 +1,351 @@
...
+                                None => return Err(format_err!("Could not find cached entries for {}.", level_name.level))
+                            };
+
+                            let neighbors_ids = match level_cache.neighbors_map.get(&cut) {
+                                Some(neighbors_ids) => neighbors_ids.clone(),
+                                None => continue
+                            };
+
+                            for neighbor_id in neighbors_ids.iter() {
+                                dimensions_map.push([level_name.level.to_string(), cut.to_string(), "neighbors".to_string(), neighbor_id.to_string()].to_vec());
```
MarcioPorto: Same as above.

tesseract-server/src/handlers/logic_layer/geoadjacents.rs
```diff
@@ -108,24 +104,20 @@ pub fn logic_layer_geoadjacent(
         None => return Ok(HttpResponse::NotFound().json("Unable to access cube cache".to_string()))
     };
 
-    let cuts_map = match agg_query.cuts{
+    let cuts_map = match agg_query.cuts {
         Some(cut_map) => cut_map,
-        None => HashMap::new(),
+        None => return Ok(HttpResponse::NotFound().json("Malformed cuts".to_string())),
```
MarcioPorto: This will be `None` when no cuts are provided, so we should specify that instead of saying they are malformatted.

tesseract-server/src/handlers/logic_layer/geoadjacents.rs
```diff
@@ -108,24 +104,20 @@ pub fn logic_layer_geoadjacent(
...
 
     let level_map = &cube_cache.level_map;
     let property_map = &cube_cache.property_map;
     let geoservice_url = &req.state().env_vars.geoservice_url;
 
-    let dimensions_map = match getdimensions(&cuts_map, &cube, &cube_cache, &level_map, &property_map, &geoservice_url) {
+    let dimensions_map: Vec<Vec<String>> = match get_relations(&cuts_map, &cube, &cube_cache, &level_map, &property_map, &geoservice_url) {
         Ok(dm) => dm,
-        Err(err) => return Ok(HttpResponse::NotFound().json(err.to_string()))
+        Err(err) => Vec::new(),
```
MarcioPorto: We should be returning any errors that come from the `get_relations` method here instead of initializing an empty vector. If your goal here is to have a way to return an empty array when there are no matches, you should return that explicitly from the `get_relations` function. That way we are still returning any possible errors.

tesseract-server/src/handlers/logic_layer/geoadjacents.rs
Jayrmajithia: I tried raising an error over her but it seems that when we don't provide the cuts value it returns `Some({})` instead of `None`. I have solved this issue by putting a check in the get_relations method

PR #231, opened 2019-11-07
Cube permissions

MarcioPorto:
Based on our discussion yesterday about features for the new OEC, this PR implements the concept of public and private cubes. Public cubes will work as normal. However, to query a private cube, a user will have to provide a valid API key in the query params or header.

This PR also simplifies error generation for multiple endpoints and fixes old compiler warnings.


tesseract-server/src/handlers/aggregate.rs
```diff
@@ -60,6 +48,15 @@ pub fn do_aggregate(
 {
     let (cube, format) = cube_format;
 
+    // Get cube object to check for API key
+    let schema = &req.state().schema.read().unwrap().clone();
+    let cube_obj = ok_or_404!(schema.get_cube_by_name(&cube));
```
jspeis: once the schema/cube_obj are unpacked here, can we reuse this for the  `..read().unwrap()..` that`query_ir_headers` does now?

jspeis: once the schema/cube_obj are unpacked here, can we reuse this for the  `..read().unwrap()..` that`query_ir_headers` does now?

jspeis:
Thanks @MarcioPorto! Overall, I think this will work as a reasonable initial protection. Eventually, I think we'd want to have some kind of federated authorization mechanism using say OAuth. Because the challenge would be if any individual shares the key (or abuses the key) then the cat's out of the bag and the key would need to change for everyone everytime that happens. But this is a good [80/20 solution](https://en.wikipedia.org/wiki/Pareto_principle)

MarcioPorto:
Exactly @jspeis! This was never meant to be the final solution, just something that works reasonably well for now. 100% agree.

PR #233, opened 2019-11-20
Adds support for the inline Unique Name for property and levels and also  alias for cube

### LINKED ISSUE #168: Add unique level names (if available) to cubes endpoint

Jayrmajithia:
closes #168


tesseract-server/src/handlers/metadata.rs
```diff
@@ -61,6 +56,131 @@ pub fn members_handler(
...
+                                            .find_unique_cube_property_name(cube_name, &property_name)
+                                            .or_else(|_| {ll_config.find_unique_shared_dimension_property_name(dimension_name, cube_name, &property_name)}){
+                                                Ok(u) => u,
+                                                Err(_) => None
+                                            };
+                                properties_list.push(PropertyMetadata{
+                                    name:property.name.clone(),
+                                    caption_set:property.caption_set.clone(),
+                                    annotations:property.annotations.clone(),
+                                    unique_name:uniq,
```
MarcioPorto: Add spaces between keys and values for readability

tesseract-server/src/handlers/metadata.rs
```diff
@@ -61,6 +56,131 @@ pub fn members_handler(
...
+                    match &level.properties {
+                        Some(p) => {
+                            for property in p.iter(){
+                                let property_name = Property::new(
+                                    &dimension.name,
+                                    &hierarchy.name,
+                                    &level.name,
+                                    &property.name,
+                                );
+                                let uniq = match ll_config
```
MarcioPorto: Use full `unique` name, or `unique_property_name` for more clarity.

jspeis:
Jumping in here for a moment. From a super quick glance it seems like there might be a bit of duplicated code between `get_cube_metadata` and `get_metadata`. I wonder if some of the common repeated steps can be refactored into a helper method? (or reduced with a macro?)

MarcioPorto:
That's a good point. I meant to check that but ended up forgetting @Jayrmajithia

tesseract-server/src/handlers/metadata.rs
```diff
@@ -61,6 +75,62 @@ pub fn members_handler(
 }
 
 
+pub fn get_cube_metadata(
+    mut cube_details: CubeMetadata,
+    ll_config: LogicLayerConfig,
```
MarcioPorto: Can we make this a reference (`&ll_config`) so we don't keep cloning it multiple times?

PR #234, opened 2019-11-21
Feature/docker integration tests

jspeis:
Initial infrastructure for end-to-end testing in using Travis+Docker. Right now it just has a simple "ping" test, but we could build more complex logic, and add a test data set from here.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -6,3 +6,4 @@
 .idea/
 bench-data
 test-schema
+.DS_Store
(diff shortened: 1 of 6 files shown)


MarcioPorto:
Thanks for setting this up @jspeis. I'll try to find some time to add that test dataset soon.

jspeis:
Sounds good, thanks @MarcioPorto that will be great to have moving forward

Issue #224, opened 2019-10-31
[logic layer] return source cube annotations in payload


davelandry:
We need this in order for the CMS to display the small source in each section (ie. "Data provided by ACS"). It doesn't _need_ to be in the same format as the node logiclayer, but something similar that passes through the annotations is needed: https://datausa.io/api/data?measure=Total%20Population&Geography=16000US3651000&year=latest

```json
{
  "data": [
    {
      "ID Year": 2017,
      "Year": "2017",
      "Total Population": 19849399,
      "Geography": "New York",
      "ID Geography": "04000US36"
    }
  ],
  "source": [
    {
      "measures": ["Total Population"],
      "annotations": {
        "source_name": "Census Bureau",
        "source_description": "The American Community Survey (ACS) Public Use Microdata Sample (PUMS) files are a set of untabulated records about individual people or housing units. The Census Bureau produces the PUMS files so that data users can create custom tables that are not available through pretabulated (or summary) ACS data products.",
        "dataset_name": "ACS PUMS 1-Year Estimate",
        "dataset_link": "https://census.gov/programs-surveys/acs/technical-documentation/pums.html",
        "subtopic": "Demographics",
        "table_id": "PUMS",
        "topic": "Diversity",
        "hidden_measures": "ygbpop RCA,ygopop RCA,ygipop RCA,yocpop RCA,yiopop RCA,ycbpop RCA"
      },
      "name": "pums_1",
      "substitutions": [
        {"id": "04000US36", "dimension": "Geography", "hierarchy": "State"}
      ]
    }
  ]
}
```

Jayrmajithia:
@davelandry below is what the output is going to look like with annotation
```
{
    "data": [
        {
            "Year": 2016,
            "Quantity": 164.0
        },
        {
            "Year": 2017,
            "Quantity": 266.0
        }
    ],
    "source": [
        {
            "name": "Sales",
            "measures": [
                "Price Total",
                "Quantity"
            ],
            "annotations": {
                "source_link1": "https://electionlab.mit.edu/data",
                "hide_in_ui": "Version",
                "dataset_name1": "1976-2016 President",
                "source_name1": "MIT Data Electionlab",
                "available_dimensions": "Candidate, Geography, Party, Year",
                "dataset_link2": "https://dataverse.harvard.edu/api/access/datafile/:persistentId?persistentId=doi:10.7910/DVN/VOQCHQ/FQ9NBF",
                "source_name2": "Federal Election Commision",
[...]

PR #237, opened 2019-12-05
Source annotations #224

Jayrmajithia:
Adds support for the Source Annotation Payload

### DIFF
--- a/tesseract-core/src/format.rs
+++ b/tesseract-core/src/format.rs
@@ -145,7 +145,7 @@ fn format_jsonrecords(headers: &[String], df: DataFrame, source_data: Option<Sou
     let mut res = String::from_utf8(ser.into_inner())?;
     let source_string = serde_json::to_string(&source_data)?;
     if source_string != "null" {
-        res.push_str(",\n\"Source\": [\n");
+        res.push_str(",\n\"source\": [\n");
         res.push_str(&source_string);
         res.push_str("\n]");
     }
--- a/tesseract-core/src/format.rs
+++ b/tesseract-core/src/format.rs
@@ -5,6 +5,7 @@ use serde::Serializer;
 use serde::ser::{SerializeSeq};
 use serde_json::{Value};
 
+use crate::schema::metadata::SourceMetadata;
 use crate::dataframe::{DataFrame, ColumnData};
 
 #[derive(Debug, Clone)]
@@ -28,10 +29,10 @@ impl std::str::FromStr for FormatType {
 }
 
 /// Wrapper to format `DataFrame` to the desired output format.
-pub fn format_records(headers: &[String], df: DataFrame, format_type: FormatType) -> Result<String, Error> {
+pub fn format_records(headers: &[String], df: DataFrame, format_type: FormatType, source_data: Option<SourceMetadata>) -> Result<String, Error> {
     match format_type {
         FormatType::Csv => Ok(format_csv(headers, df)?),
-        FormatType::JsonRecords => Ok(format_jsonrecords(headers, df)?),
+        FormatType::JsonRecords => Ok(format_jsonrecords(headers, df, source_data)?),
         FormatType::JsonArrays => Ok(format_jsonarrays(headers, df)?),
     }
 }
@@ -87,7 +88,7 @@ fn format_csv(headers: &[String], df: DataFrame) -> Result<String, Error> {
 }
 
 /// Formats response `DataFrame` to JSON records.
-fn format_jsonrecords(headers: &[String], df: DataFrame) -> Result<String, Error> {
+fn format_jsonrecords(headers: &[String], df: DataFrame, source_data: Option<SourceMetadata>) -> Result<String, Error> {
     // use streaming serializer
     // Necessary because this way we don't create a huge vec of rows containing Value
     // (very expensive)
@@ -106,8 +107,6 @@ fn format_jsonrecords(headers: &[String], df: DataFrame) -> Result<String, Error
     );
 
     let mut seq = ser.serialize_seq(Some(df.len()))?;
-
-
     // write data
     for row_idx in 0..df.len() {
         let mut row: IndexMap<&str, serde_json::Value> = IndexMap::new();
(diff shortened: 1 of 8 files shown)


tesseract-core/src/format.rs
```diff
@@ -142,9 +141,14 @@ fn format_jsonrecords(headers: &[String], df: DataFrame) -> Result<String, Error
 
         seq.serialize_element(&row)?;
     }
-
     seq.end()?;
     let mut res = String::from_utf8(ser.into_inner())?;
+    let s = serde_json::to_string(&source_data)?;
```
jspeis: lets give the `s` variable a more descriptive name

tesseract-core/src/format.rs
```diff
@@ -142,9 +141,14 @@ fn format_jsonrecords(headers: &[String], df: DataFrame) -> Result<String, Error
...
+    if s != "null"{
```
jspeis: lets add a space between the closing `"` and `{`

tesseract-server/src/handlers/aggregate.rs
```diff
@@ -91,7 +91,7 @@ pub fn do_aggregate(
         .and_then(move |df| {
             let content_type = format_to_content_type(&format);
 
-            match format_records(&headers, df, format) {
+            match format_records(&headers, df, format, None) {
```
jspeis: @Jayrmajithia can you confirm that this should this be none?

tesseract-server/src/handlers/logic_layer/aggregate.rs
```diff
@@ -335,6 +341,35 @@ pub fn logic_layer_aggregation(
 }
 
 
+// Genrates the source data/ annotaion of the cube for which the query is executed
+fn generate_source_data(cube: &Cube) -> Option<SourceMetadata> {
```
jspeis: The way this is written, this method will always return SourceMetadata, so it doesn't need to be a return type of`Option<SourceMetadata>` here. (You can wrap the response with `Some(x)`as needed?

tesseract-server/src/handlers/logic_layer/aggregate.rs
```diff
@@ -335,6 +341,35 @@ pub fn logic_layer_aggregation(
...
+            Some(anotate_hashmap)
+        },
+        None => None
+    };
+    // println!("{:?}", cube_name);
+    // println!("{:?}", measures);
+    // println!("{:?}", annotations);
+    Some(SourceMetadata {
+        name: cube_name.clone(),
+        measures: measures.clone(),
```
jspeis: I don't think we need to include the measure list here.

tesseract-server/src/handlers/logic_layer/aggregate.rs
```diff
@@ -335,6 +341,35 @@ pub fn logic_layer_aggregation(
 }
 
 
+// Genrates the source data/ annotaion of the cube for which the query is executed
+fn generate_source_data(cube: &Cube) -> Option<SourceMetadata> {
+    let cube_name = &cube.name;
+    let mut measures = Vec::new();
+    // let mut annotations = HashMap::new();
```
jspeis: can we remove this commented out line?

tesseract-server/src/handlers/aggregate.rs
Jayrmajithia: That should be none so, when we don't pass the Source data it will not be generated

tesseract-server/src/handlers/logic_layer/aggregate.rs
Jayrmajithia: Sure

tesseract-server/src/handlers/aggregate.rs
jspeis: I was thinking we might want to have source data in the response in this case because it's a data aggregation endpoint?

tesseract-core/src/schema/metadata.rs
```diff
@@ -240,3 +240,10 @@ impl From<&Aggregator> for AggregatorMetadata {
         }
     }
 }
+
+#[derive(Debug, Clone, Serialize, PartialEq)]
+pub struct SourceMetadata {
+    pub name: String,
+    pub measures: Vec<String>,
```
jspeis: we can likely eliminate the measure Vec

tesseract-server/src/handlers/aggregate.rs
Jayrmajithia: So, We have to modify the format_recode method because the data response is in format of the Dataframe. And Source will not be in the format of the Dataframe. if append in it will be added to the data property and won't create the Source property so if change the method format_recode where we pass the Source data too it will be appended in the result for the data aggregation endpoint but rest of the place where we are using the formamt_record method has to pass None

tesseract-server/src/handlers/logic_layer/aggregate.rs
jspeis: never mind, I see in Dave's examples he did include measures so we can keep this

tesseract-server/src/handlers/aggregate.rs
jspeis: Not sure I follow here? Would you be able to clarify?

tesseract-core/src/schema/metadata.rs
jspeis: never mind we will keep measures vec

tesseract-server/src/handlers/aggregate.rs
Jayrmajithia: the result from the handler/aggregate and other modules where we use this format_records method have the result in the format of Dataframe with this method it is converted to the JSON strings where the data is the only property(i.e. the output will say { data: [result]} ) present. So, when we want the source property to be added in the output result we pass the source data but if we don't want to pass the source data we just put the value None. Hope this clarifies

jspeis:
Just spoke with Jay about two changes:

1. ensuring the source metadata also gets returned when using `/aggregate` endpoints (non-logic layer endpoints)
2. lowercasing "Source" in the final JSON output to "source"

jspeis:
@Jayrmajithia I think we're getting closer. One thing I notice is that in the output it seems to produce the annotations values as a list instead of simply key-value pairs?

annotations as they appear in /cubes endpoint:
Screen Shot 2019-12-06 at 11 41 23 AM

versus the new source metadata:
Screen Shot 2019-12-06 at 11 41 38 AM


...can we have the output mirror the style from the /cubes endpoint?

jspeis:
@Jayrmajithia it looks like the function `generate_source_data` is duplicated?

tesseract-server/src/handlers/aggregate.rs
jspeis: (As a summary: we did change this behavior so that sources will also be available from non-logic layer endpoints)

MarcioPorto:
Sorry for coming in late. Looks good to me! 🚀

PR #241, opened 2019-12-10
Support the log warning if the user has write access to the data

Jayrmajithia:


Jayrmajithia:
it has been refactored now it has check_user as a trait implementation on the backend. For other databases, the idea of retrieving the SQL need to be changed.

tesseract-clickhouse/src/lib.rs
```diff
@@ -84,5 +84,31 @@ impl Backend for Clickhouse {
...
+            .get_handle()
+            .and_then(move |c| c.query(index_sql).fetch_all())
+            .and_then(move |(_, block): (_, Block<Complex>)| {
+                let values: Vec<&str> = block.get(0, "values")?;
+                let names: Vec<&str> = block.get(0, "names")?;
+                if names.contains(&"readonly") && names.contains(&"allow_ddl") {
+                    let rd_index = names.iter().position(|&r| r=="readonly").unwrap();
+                    let ad_index = names.iter().position(|&r| r=="allow_ddl").unwrap();
+                    // let rd_value = values.get(rd_index).parse::<u32>().unwrap();
+                    // let ad_value = values.get(ad_index).parse::<u32>().unwrap();
```
MarcioPorto: Remove these if they are not being used

MarcioPorto:
@jspeis Is this ready to merge?

tesseract-core/src/backend.rs
```diff
@@ -38,6 +38,10 @@ pub trait Backend {
             &query_ir.growth,
         )
     }
+
+    fn check_user(&self) -> Box<dyn Future<Item=(), Error= Error>> {
+        unimplemented!()
```
jspeis: One thing we might want to do by default here would be to log that this is unimplemented but maybe not error? thoughts?

MarcioPorto: Yeah, that's definitely better than erroring out.

MarcioPorto:
Closing in favor of https://github.com/tesseract-olap/tesseract/pull/287

PR #242, opened 2019-12-11
Ordering by RCA issue #220

### LINKED ISSUE #220: support for ordering results by RCA
The idea here is to add support for sorting results for endpoints returning RCA. Currently I am getting the following error:
```
"sort Measure(\"Trade Value RCA\") not found in measures"
```

Here is a sample URL: https://api.oec.world/tesseract/data.jsonrecords?cube=trade_i_baci_a_92&Exporter+Country=sapry&time=year.latest&drilldowns=Year&measures=Trade+Value&parents=true&rca=Exporter+Country,HS4,Trade+Value&sparse=false&limit=3&sort=Trade+Value+RCA.desc

I need this for a bar chart of top exports by RCA where i only want to show the top 10 and it would be inefficient to retrieve all results and sort them client-side.

Jayrmajithia:
Solves the issue of ordering by RCA

closes #220

### DIFF
--- a/tesseract-core/src/lib.rs
+++ b/tesseract-core/src/lib.rs
@@ -657,7 +657,7 @@ impl Schema {
             // Checking if we are going to sort by RCA or not
             let measure_name = s.measure.to_string();
             if let Some(ref rca) = query.rca {
-                if measure_name == format!("{} RCA", rca.mea.clone().to_string()){
+                if measure_name == format!("{} RCA", rca.mea.to_string()){
                     Some(SortSql {
                         direction: s.direction.clone(),
                         column: format!("rca"),
--- a/tesseract-core/src/lib.rs
+++ b/tesseract-core/src/lib.rs
@@ -655,12 +655,24 @@ impl Schema {
         let sort = if let Some(ref s) = query.sort {
             // sort column needs to be named by alias
             // Checking if we are going to sort by RCA or not
-            let measure_name = &s.measure.to_string();
-            if measure_name.contains("RCA") {
-                Some(SortSql {
-                    direction: s.direction.clone(),
-                    column: format!("rca"),
-                })
+            let measure_name = s.measure.to_string();
+            if let Some(ref rca) = query.rca {
+                if measure_name == format!("{} RCA", rca.mea.clone().to_string()){
+                    Some(SortSql {
+                        direction: s.direction.clone(),
+                        column: format!("rca"),
+                    })
+                } else {
+                    let sort_column = query.measures.iter()
+                        .position(|m| m == &s.measure)
+                        .ok_or_else(|| format_err!("sort {:?} not found in measures", &s.measure))?;
+
+
+                    Some(SortSql {
+                        direction: s.direction.clone(),
+                        column: format!("final_m{}", sort_column),
+                    })
+                }
             } else {
                 let sort_column = query.measures.iter()
                     .position(|m| m == &s.measure)
--- a/tesseract-core/src/lib.rs
+++ b/tesseract-core/src/lib.rs
@@ -656,23 +656,17 @@ impl Schema {
             // sort column needs to be named by alias
             // Checking if we are going to sort by RCA or not
             let measure_name = s.measure.to_string();
-            if let Some(ref rca) = query.rca {
-                if measure_name == format!("{} RCA", rca.mea.to_string()){
-                    Some(SortSql {
-                        direction: s.direction.clone(),
-                        column: format!("rca"),
-                    })
-                } else {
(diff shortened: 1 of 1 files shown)


tesseract-core/src/lib.rs
```diff
@@ -654,15 +654,24 @@ impl Schema {
 
         let sort = if let Some(ref s) = query.sort {
             // sort column needs to be named by alias
-            let sort_column = query.measures.iter()
-                .position(|m| m == &s.measure)
-                .ok_or_else(|| format_err!("sort {:?} not found in measures", &s.measure))?;
+            // Checking if we are going to sort by RCA or not
+            let measure_name = &s.measure.to_string();
+            if measure_name.contains("RCA") {
```
jspeis: The only thing I wonder is, should we do a more specific check here that the sort is some form of "<Column> RCA"? because now if you did `sort=Blah RCA` it would catch it even if it's not an actual column. It might just help avoid future issue in case someone has a column named like `RCAP` and they want to sort on that

Jayrmajithia: Will take a look at it

tesseract-core/src/lib.rs
```diff
@@ -655,12 +655,24 @@ impl Schema {
...
             // Checking if we are going to sort by RCA or not
-            let measure_name = &s.measure.to_string();
-            if measure_name.contains("RCA") {
-                Some(SortSql {
-                    direction: s.direction.clone(),
-                    column: format!("rca"),
-                })
+            let measure_name = s.measure.to_string();
+            if let Some(ref rca) = query.rca {
+                if measure_name == format!("{} RCA", rca.mea.clone().to_string()){
```
jspeis: do we need `clone()` & `to_string()` here? can we just use `to_string()`?

Jayrmajithia: Yep we don't need it I added by mistake

tesseract-core/src/lib.rs
```diff
@@ -655,12 +655,24 @@ impl Schema {
...
+                    Some(SortSql {
+                        direction: s.direction.clone(),
+                        column: format!("rca"),
+                    })
+                } else {
```
jspeis: would we be able to potentially avoid duplicating this `else` block by converting the nested `if`s into a single compound if statement e.g. `if X && Y ` instead of `if X { if Y { ...` ... ?

tesseract-core/src/lib.rs
```diff
@@ -657,7 +657,7 @@ impl Schema {
             // Checking if we are going to sort by RCA or not
             let measure_name = s.measure.to_string();
             if let Some(ref rca) = query.rca {
-                if measure_name == format!("{} RCA", rca.mea.clone().to_string()){
+                if measure_name == format!("{} RCA", rca.mea.to_string()){
```
jspeis: since this is in the context of a format! , do we even need to the to_string()? Might be able to remove that too?

tesseract-core/src/lib.rs
Jayrmajithia: We cannot because I need to get the column name of mea inside the RCA

tesseract-core/src/lib.rs
Jayrmajithia: We need to_string because the rca.mea is of format measure

tesseract-core/src/lib.rs
jspeis: In that case, could we maybe rewrite the `if` into a match?      like      ```match query.rca {
                Some(ref rca) => {

                },
                None =>
            }``` could that capture the flow while avoiding the duplication?

Jayrmajithia: But we still need to duplicate it in a scenario when we have rca column and we perform a sort on something else that is not in the RCA column it won't work. The fix could be we Just first check if the sort.measure contains the  contains RCA in it or not and if contains we do a strict check

jspeis: I guess I'm really averse to extra nesting and duplication so I'm trying to find a way around this :) thoughts about an approach like this?

```rust
            let rca_mea_name = if query.rca.is_some() {
                 query.rca.as_ref().unwrap().mea.to_string() 
            } else {
                "".to_string()
            };
            if measure_name == format!("{} RCA", rca_mea_name) {
                    Some(SortSql {
                        direction: s.direction.clone(),
                        column: "rca".to_string(),
                    })
            }
```

Does this work? What are your thoughts?

Jayrmajithia: I like this idea because it has less complexity of nesting. This should work

Issue #229, opened 2019-11-05
Locales don't work when RCA is requested


MarcioPorto:

MarcioPorto:
Updating the description of this ticket, since it works in the logic layer: https://api.oec.world/tesseract/data.jsonrecords?cube=trade_i_baci_a_92&drilldowns=Exporter+Country&locale=es&measures=Trade+Value&parents=false&rca=Year%2CHS2%2CTrade+Value&sparse=false&Exporter+Country=sabra.

Turns out that raw Tesseract does not currently have support for a `locale` param.

PR #244, opened 2019-12-12
Locale name of levels  in RCA Solves the issue #229 

Jayrmajithia:
Adds support to get the locale names of the levels in RCA using the logic layer


jspeis:
@Jayrmajithia looking at #229 that seems to refer to an issue in raw tesseract and not the logic layer, but this PR appears to fix an issue within the logic layer, is this part of another issue?

Jayrmajithia:
Actually, we found after changing the name that the level name provided in the RCA where never being checked if there was any change at the logic layer it worked on the levels that are present in the drill-down and cuts. And we forgot to edit the name of the issue again

PR #246, opened 2019-12-16
Logic Layer improvements

Jayrmajithia:
1) Added Support to perform filters query from Logic Layer.
2) Fixed the sorting bug on the measure used in RCA.
3) Adds support to perform multiple filters operations(i.e. filters=measure1.comparison1.value1, measure2.comparison2.value2).

### DIFF
--- a/tesseract-clickhouse/src/sql.rs
+++ b/tesseract-clickhouse/src/sql.rs
@@ -12,13 +12,13 @@ use tesseract_core::query_ir::{
     DrilldownSql,
     MeasureSql,
     HiddenDrilldownSql,
-    TopSql,
-    TopWhereSql,
-    SortSql,
-    LimitSql,
+    // TopSql,
+    // TopWhereSql,
+    // SortSql,
+    // LimitSql,
     RcaSql,
     GrowthSql,
-    FilterSql,
+    // FilterSql,
     RateSql,
     dim_subquery,
 };
@@ -144,4 +144,3 @@ mod test {
     }
 
 }
-
--- a/tesseract-clickhouse/src/sql.rs
+++ b/tesseract-clickhouse/src/sql.rs
@@ -12,13 +12,8 @@ use tesseract_core::query_ir::{
     DrilldownSql,
     MeasureSql,
     HiddenDrilldownSql,
-    // TopSql,
-    // TopWhereSql,
-    // SortSql,
-    // LimitSql,
     RcaSql,
     GrowthSql,
-    // FilterSql,
     RateSql,
     dim_subquery,
 };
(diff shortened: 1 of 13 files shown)


jspeis:
For #2 @Jayrmajithia would you be able to add a test case that confirms the fix? And for #3 could we add a test case for this new functionality as well?

Jayrmajithia:
Still working on adding test case in the same PR and also will be pushing to support the `and` and `or` in filter query

tesseract-clickhouse/src/sql/options.rs
```diff
@@ -76,7 +76,14 @@ pub fn wrap_options(
 
     let mut filters_sql = if !filters.is_empty() {
         let filter_clauses = filters.iter()
-            .map(|f| format!("{} {}", f.by_column, f.constraint.sql_string()));
+            .map(|f| {
+                if let Some(o) = &f.operator {
+                    let c2 = f.constraint2.as_ref().unwrap();
+                    format!("({} {} {} {} {})", f.by_column, f.constraint.sql_string(), o.sql_string(), f.by_column, c2.sql_string())
```
jspeis: can we make the o / f /c2 variables more explicit in terms of what they are meant to represent?

Jayrmajithia:
Also solves the issue # 201 few warning left are because of the logic which will raise an error if made changes in those warnings

tesseract-clickhouse/src/sql.rs
```diff
@@ -12,13 +12,13 @@ use tesseract_core::query_ir::{
...
     MeasureSql,
     HiddenDrilldownSql,
-    TopSql,
-    TopWhereSql,
-    SortSql,
-    LimitSql,
+    // TopSql,
+    // TopWhereSql,
+    // SortSql,
+    // LimitSql,
```
jspeis: we should delete these lines if not longer needed?

tesseract-clickhouse/src/sql.rs
```diff
@@ -12,13 +12,13 @@ use tesseract_core::query_ir::{
...
     RcaSql,
     GrowthSql,
-    FilterSql,
+    // FilterSql,
```
jspeis: see above

tesseract-core/src/query.rs
```diff
@@ -425,30 +425,96 @@ impl FromStr for GrowthQuery {
...
+
+                    Ok(FilterQuery {
+                        by_mea_or_calc,
+                        constraint,
+                        operator: None,
+                        constraint2: None
+                    })
+                },
+                _ => bail!("Could not parse a filter query"),
+            }
```
jspeis: this might be simplified with a regular expression?

Jayrmajithia: It can be solved using the regex or also without regex and in a simpler format than what I have done, I would love to contribute if that is okay.

Jayrmajithia: @jspeis just made a commit solving this issue just make sure to test it before merging it the logic works it should not have any issues. I was not able to run the test because of a lack of Openssl in the windows system and also I don't have a local clickhouse set up in my current system as of right now. Let me know if any other changes are needed

jspeis: Wow, great! Thanks for going above & beyond Jay! I'll take a look

PR #251, opened 2019-12-19
Fix for logic layer string measures

### LINKED ISSUE #240: measures returned as strings when data has null values?

MarcioPorto:
closes #240

### DIFF
--- a/tesseract-server/src/handlers/logic_layer/aggregate.rs
+++ b/tesseract-server/src/handlers/logic_layer/aggregate.rs
@@ -264,7 +264,6 @@ pub fn logic_layer_aggregation(
                 if same_type {
                     let mut column_data: ColumnData = ColumnData::Text(col_data.clone());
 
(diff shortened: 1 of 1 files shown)


jspeis:
Two quick thoughts:

1. could we macro-ize:

```rust
ColumnData::NullableInt16(col_data.iter().map(|x| {
                                if x == "" {
                                    None
                                } else {
                                    Some(x.parse::<i16>().unwrap())
                                }
                            }).collect());
``` to instead something like `read_data(ColumnData::NullableInt16, i16)` ? (not sure if possible to pass a type to a macro? but if so could be cool?)

only other thing what do you think about using expect vs. unwrap with after the parsing? just thinking if that might might make bug hunting a little easier in case a panic happens

MarcioPorto:
Totally agree with your suggestions @jspeis. Here's a refactored version of this functionality.

PR #255, opened 2020-01-10
adds Dockerfile for docker distribution

frabarz:


Dockerfile
```diff
@@ -0,0 +1,15 @@
...
+
+FROM debian:buster-slim
+RUN groupadd -r tesseract &&\
+    useradd -r -g tesseract tesseract &&\
+    apt-get update &&\
+    apt-get install --no-install-recommends -y libssl-dev &&\
+    rm -rf /var/lib/apt/lists/*
+COPY --from=builder /usr/local/cargo/bin/tesseract-olap /usr/local/bin/tesseract-olap
+USER tesseract
+EXPOSE 7000
```
MarcioPorto: Should we expose 7777 for consistency?

frabarz: good point

Dockerfile
```diff
@@ -0,0 +1,15 @@
+FROM rust:1.39 as builder
```
jspeis: is there a way we can use latest rust stable?

frabarz: of cors

jspeis:
@frabarz would `rust:latest` work? https://hub.docker.com/_/rust/
Screen Shot 2020-01-26 at 9 20 34 AM

frabarz:
well, yeah, but I chose to use a specific version to ensure stability, since we can't never be sure on future versions
should I use `latest`?

jspeis:
if latest == stable ... then we should be ok because the current Travis CI tool will test to make sure the build works on stable rust https://github.com/tesseract-olap/tesseract/blob/master/.travis.yml#L3

jspeis:
(otherwise 1.40 works!)

frabarz:
Just read the docs for Travis CI, and for `stable` they use the latest version available through `rustup`, so I'm ok with using `latest`.
But I also read something interesting:
image
I think it's worth adding at least `beta` too, so we are prepared and know ahead of any breaking change. Tesseract is growing, and we are now maintaining more distribution channels, so preventing issues is good for the management.

PR #269, opened 2020-02-03
Fixes name resolution bug

MarcioPorto:


tesseract-server/src/logic_layer/cache.rs
```diff
@@ -236,14 +236,21 @@ impl CubeCache {
     // non-logic layer setups, but also be used for members endpoint (which requires label also)
     pub fn members_for_level(&self, level_name: &LevelName) -> Option<&HashSet<String>> {
         debug!("Level Caches: {:?}", self.level_caches);
-        self.level_caches.get(&level_name.level)
-            .map(|level_cache| &level_cache.members)
+
+        for (_key, value) in self.level_caches.iter() {
+            if value.level_name == level_name.clone() {
```
jspeis: Why is the value_level name different than the key? Does it need to be?

MarcioPorto: Good call. Just pushed a change that makes the `LevelName` object the key

jspeis:
@MarcioPorto any idea why the build is failing? Do you tests work locally?

PR #270, opened 2020-02-06
Adds finer grained JWT access control

jspeis:
* Users can now give cubes a `minAuthLevel` value. This will require that tokens must have at least that level or higher in the `auth_level` claim of their token. By default valid users without an explicit auth level will have an auth level of 0, and also cubes will have a default auth level of 0.
* The `/cubes` endpoint listing will now filter out cubes that a user does not have access to.

### DIFF
--- a/tesseract-core/src/schema.rs
+++ b/tesseract-core/src/schema.rs
@@ -127,19 +127,16 @@ impl From<SchemaConfigJson> for Schema {
 
             // Cubes are public by default
             let public = match cube_config.public {
-                Some(public) => {
-                    if public == "false" {
-                        false
-                    } else {
-                        true
-                    }
-                },
+                Some(public) => public != "false",
                 None => true
             };
+
+            let min_auth_level = cube_config.min_auth_level.unwrap_or(0);
 
             cubes.push(Cube {
                 name: cube_config.name,
                 public,
+                min_auth_level,
                 table: cube_config.table.into(),
                 can_aggregate: false,
                 dimensions,
@@ -170,6 +167,7 @@ impl From<SchemaConfigJson> for Schema {
 pub struct Cube {
     pub name: String,
     pub public: bool,
+    pub min_auth_level: i32,
     pub table: Table,
     pub can_aggregate: bool,
     pub dimensions: Vec<Dimension>,
@@ -790,6 +788,7 @@ mod test {
                 CubeConfigJson {
                     name: "test_cube".into(),
                     public: Some("true".into()),
+                    min_auth_level: None,
                     table: TableConfigJson {
                         name: "fact_table".into(),
                         schema: None,
--- a/tesseract-core/src/schema.rs
+++ b/tesseract-core/src/schema.rs
@@ -34,7 +34,7 @@ pub use crate::schema::{
 use crate::names::{LevelName, Measure as MeasureName, Property as TsProperty};
 use crate::query_ir::MemberType;
 pub use self::aggregator::Aggregator;
-
+use crate::DEFAULT_ALLOWED_ACCESS;
 
 #[derive(Debug, Clone, PartialEq, Serialize, Deserialize)]
 pub struct Schema {
@@ -131,7 +131,7 @@ impl From<SchemaConfigJson> for Schema {
                 None => true
             };
 
-            let min_auth_level = cube_config.min_auth_level.unwrap_or(0);
+            let min_auth_level = cube_config.min_auth_level.unwrap_or(DEFAULT_ALLOWED_ACCESS);
 
             cubes.push(Cube {
                 name: cube_config.name,
--- a/tesseract-core/src/schema/xml.rs
+++ b/tesseract-core/src/schema/xml.rs
@@ -33,6 +33,8 @@ pub struct SchemaConfigXML {
 pub struct CubeConfigXML {
     pub name: String,
     pub public: Option<String>,
+    #[serde(rename(deserialize="minAuthLevel"))]
+    pub min_auth_level: Option<i32>,
     #[serde(rename(deserialize="Table"))]
     pub table: TableConfigXML,
     #[serde(rename(deserialize="Dimension"))]
--- a/tesseract-core/src/schema/xml.rs
+++ b/tesseract-core/src/schema/xml.rs
@@ -33,7 +33,7 @@ pub struct SchemaConfigXML {
 pub struct CubeConfigXML {
     pub name: String,
     pub public: Option<String>,
-    #[serde(rename(deserialize="minAuthLevel"))]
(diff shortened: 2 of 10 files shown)


tesseract-server/src/auth.rs
```diff
@@ -55,21 +31,76 @@ struct Claims {
...
-                Err(_) => false, // If any error occurs, do not validate
+                Err(_err) => Some(-1), // If any error occurs, do not validate
+            }
+        },
+        None => None
+    }
+}
+
+pub fn validate_web_token(jwt_secret: &Option<String>, raw_token: &str, min_auth_level: i32) -> bool {
+    match jwt_secret {
```
MarcioPorto: Should we log a message on startup saying that the JWT secret hasn't been set and what that means? I think the reasoning here is the same as logging the read/write access for the database user. It's a way of making sure this is what the Tesseract user intended to do.

tesseract-server/src/auth.rs
```diff
@@ -55,21 +31,76 @@ struct Claims {
...
+        Some(key) => {
+            let validation = Validation::default();
+            match decode::<Claims>(&raw_token, key.as_ref(), &validation) {
+                Ok(c) => {
+                    let claims: Claims = c.claims;
+                    let part1 = claims.status == "valid"; // TODO allow this value to be configurable
+                    let part2 = match claims.auth_level {
+                        Some(lvl) => lvl >= min_auth_level,
+                        None => true
```
MarcioPorto: Should we require that an `auth_level` be provided? I believe that right now if no level was included in the JWT, we'd allow cube access even if that cube had a min auth level > 0, no?

tesseract-server/src/auth.rs
```diff
@@ -55,21 +31,76 @@ struct Claims {
...
+            let validation = Validation::default();
             match decode::<Claims>(&raw_token, key.as_ref(), &validation) {
                 Ok(c) => {
                     let claims: Claims = c.claims;
-                    claims.status == "valid" // TODO allow this value to be configurable
+                    let part1 = claims.status == "valid";
+                    if part1 {
+                        Some(claims.auth_level.unwrap_or(0)) // default auth level 0
+                    } else {
+                        Some(0)
```
MarcioPorto: Should we refer to the `DEFAULT_ALLOWED_ACCESS` variable instead of using a literal 0 here instead?

tesseract-server/src/handlers/metadata.rs
```diff
@@ -44,14 +44,26 @@ pub fn metadata_all_handler(
...
-        cubes.push(get_cube_metadata(cube.clone(), &ll_config));
+        // Filter out cube that user isn't authorized to see
+        match user_auth_level {
+            Some(auth_level) => { // Authorization is set
+                if (auth_level >= cube.min_auth_level && auth_level >= 0) {
+                    cubes.push(get_cube_metadata(cube.clone(), &ll_config));
+                }
+            },
+            // No authorization set. Show all cubes
+            None => cubes.push(get_cube_metadata(cube.clone(), &ll_config))
```
MarcioPorto: Similar to what I mentioned above, I think that when no level is provided, only the cubes with min_level = 0 should be returned.

tesseract-server/src/handlers/metadata.rs
```diff
@@ -44,14 +44,26 @@ pub fn metadata_all_handler(
...
+            return  Ok(HttpResponse::Ok().json(schema_details))
+        }
     };
     let mut cubes: Vec<CubeMetadata> = Vec::new();
     for cube in schema_details.cubes.iter(){
-        cubes.push(get_cube_metadata(cube.clone(), &ll_config));
+        // Filter out cube that user isn't authorized to see
+        match user_auth_level {
+            Some(auth_level) => { // Authorization is set
+                if (auth_level >= cube.min_auth_level && auth_level >= 0) {
```
MarcioPorto: Use `DEFAULT_ALLOWED_ACCESS` variable instead?

tesseract-server/src/handlers/util.rs
```diff
@@ -70,7 +70,11 @@ pub fn generate_source_data(cube: &Cube) -> SourceMetadata {
...
 }
 
-
+pub fn get_user_auth_level(req: &HttpRequest<AppState>) -> Option<i32> {
+    let jwt_secret = &req.state().env_vars.jwt_secret;
+    let user_token = extract_token(req);
+    user_auth_level(jwt_secret, &user_token)
+}
 
 pub fn verify_api_key(req: &HttpRequest<AppState>, cube: &Cube) -> Result<(), HttpResponse> {
```
MarcioPorto: Just to confirm, we want to keep the old check for public cubes AND also add the new `min_auth_level` check? I had initially thought that we would only be using the latter, but I guess it makes sense to keep the public flag for applications that won't use JWTs

tesseract-core/src/schema.rs
```diff
@@ -127,19 +127,16 @@ impl From<SchemaConfigJson> for Schema {
...
-                        false
-                    } else {
-                        true
-                    }
-                },
+                Some(public) => public != "false",
                 None => true
             };
 
+            let min_auth_level = cube_config.min_auth_level.unwrap_or(0);
```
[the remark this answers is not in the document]
jspeis: Yup!

tesseract-server/src/auth.rs
jspeis: On startup that's definitely reasonable to add a log

tesseract-server/src/auth.rs
jspeis: Yup

tesseract-core/src/schema/xml.rs
```diff
@@ -33,6 +33,8 @@ pub struct SchemaConfigXML {
 pub struct CubeConfigXML {
     pub name: String,
     pub public: Option<String>,
+    #[serde(rename(deserialize="minAuthLevel"))]
```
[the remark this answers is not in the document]
jspeis: 👍  I'll change that. I would have thought that in XML could be nice to follow a camel case convention but I agree it makes sense to stick to go with what already are using

tesseract-server/src/auth.rs
jspeis: At first I was thinking it would be easier to allow default access level if no auth level was specified to be 0 but I think it in terms of security probably better to err on the side of requiring more specificity than less, so I'll change this

tesseract-server/src/handlers/metadata.rs
jspeis: this is a case though where the server is not using JWT _at all_ ... because we don't want to necessarily force it. In case the JWT is set but auth is invalid it will be Some(-1)

PR #273, opened 2020-02-20
Adds unique header names when appropriate

MarcioPorto:
Right now unique level names are not being returned as the header in the response: https://api.oec.world/tesseract/data.jsonrecords?cube=trade_i_baci_a_92&drilldowns=Exporter+Continent%2CImporter+Continent&measures=Trade+Value&parents=false&sparse=false

This PR fixes that by substituting the header names for the unique version whenever there are multiple headers with the same name.

Sample response from the same query:

Screen Shot 2020-02-20 at 1 38 36 PM

### DIFF
--- a/tesseract-server/src/handlers/logic_layer/aggregate.rs
+++ b/tesseract-server/src/handlers/logic_layer/aggregate.rs
@@ -223,8 +223,40 @@ pub fn logic_layer_aggregation(
         return boxed_error_string("Unable to generate queries".to_string())
     }
 
+    // Need to create a map here to help create unique header names in the next step
+    let mut unique_header_map: HashMap<String, String> = HashMap::new();
+
+    if let Some(ref llc) = logic_layer_config {
+        if let Some(ref llc_aliases) = llc.aliases {
+            if let Some(ref llc_cubes) = llc_aliases.cubes {
+                for llc_cube in llc_cubes {
+                    if cube_name == llc_cube.name {
+                        if let Some(ref llc_cube_levels) = llc_cube.levels {
+                            for llc_cube_level in llc_cube_levels {
+                                unique_header_map.insert(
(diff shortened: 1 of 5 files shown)


jspeis:
@MarcioPorto seems like the build fails?

MarcioPorto:
@jspeis Should be all good now.

tesseract-server/src/handlers/logic_layer/aggregate.rs
```diff
@@ -223,8 +223,40 @@ pub fn logic_layer_aggregation(
...
+                                    llc_cube_property.current_name.clone(),
+                                    llc_cube_property.unique_name.clone()
+                                );
+                            }
+                        }
+                    }
+                }
+            }
+        }
+    }
```
jspeis: Just thinking out loud here, seeing the deep nesting here makes me wonder if it could be possible to delegate some of the complexity to trait functions? (e.g. in `LogicLayerConfig`, `AliasConfig` etc?)

For instance what if LLC had a trait method that basically did the equivalent of:

```rust
    if let Some(ref llc) = logic_layer_config {
        if let Some(ref llc_aliases) = llc.aliases {
            if let Some(ref llc_cubes) = llc_aliases.cubes {
...
```
so if you could say:
```
let llc_cubes = logic_layer_config.do_magic_here();
```

might help abstract some of the complexity/code nesting?

MarcioPorto: @jspeis What if I just move the whole method to `LogicLayerConfig`?

Moving the first few `if let Some` statements would only decrease the indent by two because you still need the first of those statements to check that `logic_layer_config` is not `None`.

jspeis: I'm ok with that for now, and perhaps in the future we can revisit

PR #274, opened 2020-02-24
First Pass at Data QA/Diagnosis

MarcioPorto:


jspeis:
can the error reporting be added for streaming responses too?

tesseract-server/src/handlers/diagnosis.rs
```diff
@@ -0,0 +1,174 @@
...
+    // Check for `MissingDimensionIDs`
+    // TODO: Deal with the case where there is an inline table.
+    for dimension in &cube.dimensions {
+        for hierarchy in &dimension.hierarchies {
+            let last_level: &Level = &hierarchy.levels[hierarchy.levels.len() - 1];
+
+            if let Some(ref foreign_key) = dimension.foreign_key {
+                if let Some(ref dimension_table) = hierarchy.table {
+                    let sql_str: String = format!(
+                        "SELECT DISTINCT {} FROM {} WHERE {} NOT IN (SELECT {} FROM {})",
```
jspeis: Two things: 
1. Just to double check: nothing in the inputs here is coming from the user? It's all only from schema? Just thinking about SQL injection
2. For now I think this is OK, though future because this kind of query should work in most SQL variants AFAIK, though we might want to delegate this to the backend driver in the future

MarcioPorto: 1. Yeah, the only thing we get from the user is the cube name.
2. Very good point. I think is generic enough that it would work for most, if not all, variants we support. But as we add more checks we should definitely refactor this.

PR #276, opened 2020-03-03
Authorization refactor

jspeis:
* removes old API key verification.
* reworks JWT auth to allow access by default. (All requests will be considered to have an `auth_level` of 0)

### DIFF
--- a/tesseract-server/src/handlers/util.rs
+++ b/tesseract-server/src/handlers/util.rs
@@ -42,10 +42,6 @@ pub fn boxed_error_http_response(response: HttpResponse) -> FutureResponse<HttpR
     Box::new(future::result(Ok(response)))
 }
 
-
-pub const X_TESSERACT_API_KEY: &str = "x-tesseract-api-key";
-
-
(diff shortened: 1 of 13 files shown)


tesseract-server/src/handlers/util.rs
```diff
@@ -42,10 +42,6 @@ pub fn boxed_error_http_response(response: HttpResponse) -> FutureResponse<HttpR
     Box::new(future::result(Ok(response)))
 }
 
-
-pub const X_TESSERACT_API_KEY: &str = "x-tesseract-api-key";
```
MarcioPorto: Since we're removing this, we can also remove line 92 from `tesseract-server/src/main.rs` as well as the `api_key` object from the `EnvVars` struct in `tesseract-server/src/app.rs`

jspeis: good catch! I'll remove those

jspeis: @MarcioPorto I've just pushed those removals

PR #281, opened 2020-04-14
Logic Layer support for "exclude"

MarcioPorto:
Adds support for a new "exclude" param like this: `exclude=Subnat+Geography:48,49;Trade+Flow:1`. 

Excluding multiple levels is also supported, as long as they are separated by a `;` as in the example above. Since this is the logic layer, unique level names are expected.

### DIFF
--- a/tesseract-server/src/handlers/logic_layer/aggregate.rs
+++ b/tesseract-server/src/handlers/logic_layer/aggregate.rs
@@ -440,7 +440,7 @@ pub fn logic_layer_aggregation(
                 // multiple data types. In those cases, we will convert the
                 // whole column to string values.
                 if same_type {
-                    let mut column_data: ColumnData = ColumnData::Text(col_data.clone());
+                    let mut column_data: ColumnData = ColumnData::Text(vec![]);
 
                     match first_col.column_data {
                         ColumnData::Int8(_) => {
@@ -512,7 +512,9 @@ pub fn logic_layer_aggregation(
                                 }
                             }).collect());
                         }
-                        _ => ()
+                        _ => {
+                            column_data = ColumnData::Text(col_data.clone());
+                        }
                     }
 
                     final_columns.push(Column {
--- a/tesseract-server/src/handlers/logic_layer/aggregate.rs
+++ b/tesseract-server/src/handlers/logic_layer/aggregate.rs
@@ -440,82 +440,80 @@ pub fn logic_layer_aggregation(
                 // multiple data types. In those cases, we will convert the
                 // whole column to string values.
                 if same_type {
-                    let mut column_data: ColumnData = ColumnData::Text(vec![]);
-
-                    match first_col.column_data {
+                    let column_data = match first_col.column_data {
                         ColumnData::Int8(_) => {
-                            column_data = ColumnData::Int8(consolidate_column_data!(&col_data, i8));
+                            ColumnData::Int8(consolidate_column_data!(&col_data, i8))
                         },
                         ColumnData::Int16(_) => {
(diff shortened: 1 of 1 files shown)


tesseract-server/src/handlers/logic_layer/aggregate.rs
```diff
@@ -294,6 +334,85 @@ pub fn logic_layer_aggregation(
...
+            for col_i in 0..num_cols {
+                let mut col_data: Vec<String> = vec![];
+
+                for df in &dfs {
+                    let c: &Column = &df.columns[col_i];
+                    let rows = c.stringify_column_data();
+                    col_data = [&col_data[..], &rows[..]].concat()
+                }
+
+                // Find rows that need to be excluded
```
jspeis, line 363: would it make more sense to have the exclusion happen in the query itself rather than return the rows and then filter them out?

tesseract-server/src/handlers/logic_layer/aggregate.rs
```diff
@@ -130,6 +131,43 @@ impl LogicLayerQueryOpt {
 
         arg_vec
     }
+
+    pub fn deserialize_exclude(&self) -> HashMap<String, HashSet<String>> {
+        let mut excludes: HashMap<String, HashSet<String>> = HashMap::new();
+
+        match &self.exclude {
+            Some(arg) => {
+                let levels: Vec<String> = arg.split(";").map(|s| s.to_string()).collect();
```
jspeis, line 140: can we just use reference here? e.g. 
```rust
let levels: Vec<&str> = arg.split(";").collect();
```

tesseract-server/src/handlers/logic_layer/aggregate.rs
```diff
@@ -130,6 +131,43 @@ impl LogicLayerQueryOpt {
...
+
+                for level in levels {
+                    let parts: Vec<String> = level.split(":").map(|s| s.to_string()).collect();
```
jspeis, line 143: same here?

tesseract-server/src/handlers/logic_layer/aggregate.rs
```diff
@@ -130,6 +131,43 @@ impl LogicLayerQueryOpt {
...
+                        Some(level_name) => level_name,
+                        None => continue
+                    };
+
+                    let level_ids = match parts.get(1) {
+                        Some(level_ids) => level_ids,
+                        None => continue
+                    };
+
+                    let level_ids: Vec<String> = level_ids.split(",").map(|s| s.to_string()).collect();
```
jspeis, line 155: same?

tesseract-server/src/handlers/logic_layer/aggregate.rs
```diff
@@ -429,88 +429,82 @@ pub fn logic_layer_aggregation(
...
+                    .map(|(_, e) | e.to_string())
+                    .collect();
 
                 // When returning data from multiple levels from the same
                 // hierarchy, there is a chance that this column will have
                 // multiple data types. In those cases, we will convert the
                 // whole column to string values.
                 if same_type {
-                    let mut column_data: ColumnData = ColumnData::Text(final_col_data.clone());
+                    let mut column_data: ColumnData = ColumnData::Text(col_data.clone());
```
jspeis: Would it make sense to change this block so that instead of having `let mut column_data: ColumnData = ColumnData::Text(col_data.clone());` we did something like `let column_data = match first_col.column_data {...` ? in the current case it looks like we might clone `col_data` and store that in column_data, even if we end up overwriting `column_data`?

PR #287, opened 2020-05-01
Default readonly

D2SD:


tesseract-clickhouse/src/lib.rs
```diff
@@ -22,18 +24,34 @@ pub struct Clickhouse {
...
+
         let options = format!("tcp://{}", url).parse::<Options>()?;
 
         let options = options
             // Ping timeout is necessary, because under heavy load (100 requests
             // simultaneously, each one taking 5s ordinarily) the client will timeout.
             .ping_timeout(Duration::from_millis(PING_TIMEOUT));
 
-        let pool = Pool::new(options);
+        if rg.captures(url).is_none() {
```
jspeis, line 36: I think a simpler way to express this might be: 

        let options = options
            .readonly(match rg.captures(url) { 
                Some(_) => Some(1),
                None => None
            })
            .ping_timeout(Duration::from_millis(PING_TIMEOUT));
This way we might reduce some code duplication

jspeis:
seems like the integration test is failing, have you had a chance to test locally?

tesseract-clickhouse/src/lib.rs
D2SD: since `rg.captures(url)` returns `true|false` i have to do this in order to get the value of the matched regex 

```
        let options = options
            .readonly(match rg.captures(url) { 
                Some(_) =>  Some(rg.captures(url).unwrap().name("id")
                                .unwrap().as_str().parse::<u8>().unwrap()),
                None => Some(1)
            })
            .ping_timeout(Duration::from_millis(PING_TIMEOUT));
```

ran the test locally and the error was that it didn't found the table that was using for the tests

D2SD: after digging a bit, the test failed at `tesseract-clickhouse/tests/integration.rs` the `test_query`, 
```
---- test_query stdout ----
Value=[RowResult { month_name: "July" }, RowResult { month_name: "April" }, RowResult { month_name: "December" }, RowResult { month_name: "January" }, RowResult { month_name: "July" }, RowResult { month_name: "December" }, RowResult { month_name: "April" }, RowResult { month_name: "August" }, RowResult { month_name: "January" }, RowResult { month_name: "May" }, RowResult { month_name: "February" }, RowResult { month_name: "October" }, RowResult { month_name: "August" }, RowResult { month_name: "May" }, RowResult { month_name: "October" }, RowResult { month_name: "February" }, RowResult { month_name: "June" }, RowResult { month_name: "September" }, RowResult { month_name: "November" }, RowResult { month_name: "March" }, RowResult { month_name: "November" }, RowResult { month_name: "September" }, RowResult { month_name: "June" }, RowResult { month_name: "March" }]
```
since the response is longer than 12 elements it triggers the error
```
        Ok(val) => {
            println!("Value={:?}", val);
            val.len() == 12
        }
        Err(err) => {
            println!("Query failed. Error={:?}", err);
            false
        }
```

D2SD: it seems it does not recognize the `readonly=1` param
`thread '<unnamed>' panicked at called Result::unwrap() on an Err value: ErrorMessage { msg: "Cache population error: Error populating cache with backend data: Server error: ERROR DB::Exception (115): DB::Exception: Unknown setting 1" }', tests/src/clickhouse_end_to_end.rs:169:25`

jspeis: I think in the end this may be caused by an issue with how Travis-CI handles PRs from forks (I think it was running the tests on the wrong target branch)

tesseract-clickhouse/src/lib.rs
```diff
@@ -22,11 +24,19 @@ pub struct Clickhouse {
...
 
         let options = options
-            // Ping timeout is necessary, because under heavy load (100 requests
-            // simultaneously, each one taking 5s ordinarily) the client will timeout.
+            .readonly(match rg.captures(url) {
+                Some(readonly_option) => {
+                    let rg_match = readonly_option.name("id").expect("Could not parse a value for readonly").as_str();
+                    let num_match = rg_match.parse::<u8>().expect(&format!("Failed to parse {} into a numeric value", rg_match));
+                    Some(num_match)
+                            }
```
jspeis, line 37: can we fix the indentation for this bracket?

PR #288, opened 2020-05-04
Compiler warnings cleanup

D2SD:


tesseract-core/src/lib.rs
```diff
@@ -1257,7 +1257,7 @@ impl Schema {
         Ok(column)
     }
 
-    fn get_mea_col(&self, cube_name: &str, measure: &Measure) -> Result<String, Error> {
+    /* fn get_mea_col(&self, cube_name: &str, measure: &Measure) -> Result<String, Error> {
```
jspeis, line 1260: instead of commenting this out, lets removing this method if its not used

PR #290, opened 2020-05-11
deduplicate drilldown

D2SD:


tesseract-server/src/handlers/aggregate.rs
```diff
@@ -69,7 +81,9 @@ pub fn do_aggregate(
         static ref QS_NON_STRICT: qs::Config = qs::Config::new(5, false);
     }
     let agg_query_res = QS_NON_STRICT.deserialize_str::<AggregateQueryOpt>(&query);
-    let agg_query = ok_or_404!(agg_query_res);
+    let mut agg_query = ok_or_404!(agg_query_res);
+
+    agg_query.drilldowns = Some(deduplicate_vec(&agg_query.drilldowns.expect("Could not convert to vector")));
```
jspeis, line 86: what do you think about doing this here versus in the impl of AggregateQueryOpt? Also we may be able to use `unique` method from [itertools](https://stackoverflow.com/questions/47636618/vecdedup-does-not-work-how-do-i-deduplicate-a-vector-of-strings) for deduplication ?  (or sort + dedup)

PR #298, opened 2021-01-13
adds basic name filtering on member endpoints

frabarz:


tesseract-server/src/handlers/metadata.rs
```diff
@@ -198,7 +207,38 @@ pub fn do_members(
 }
 
 
+pub fn apply_filter_term(df: DataFrame, term: &str) -> DataFrame {
+    debug!("Filtering dataframe by: {}", term);
+
+    // Collect all column names with text data
+    let columns: Vec<String> = (&df.columns).iter()
+        .filter_map(|column| match column.column_data {
+            ColumnData::Text(_) => Some(column.name.clone()),
```
hwchen, line 216: I think you can use a `matches!` macro here if you like

hwchen:
@frabarz thanks for this PR!

I think I understand the overall gist, although a few sentences of explain would still be helpful.

The parts I want to know more about:
- `drain_filter` I think works a little differently than the one in std but appears to be similar in result. https://doc.rust-lang.org/std/vec/struct.Vec.html#method.drain_filter . I didn't look too closely, but is there a reason you didn't use the one in std? I assume that one should be more efficient since they can use unsafe internally.
- Can you document a bit the place where `drain_filter` is used in the handler?

And just for clarification: this PR is so that when asking for members, can filter by some pattern/regex. And this is needed because the endpoint may return a large number of members the user doesn't care about.

tesseract-core/src/dataframe.rs
```diff
@@ -20,35 +21,65 @@ impl DataFrame {
...
+                        None => Datum::NullableText(None),
+                    },
+                };
+
+                datum.insert(&column.name, value);
+            }
+
+            if !predicate(datum, i) {
+                for column in self.columns.iter_mut() {
+                    column.remove(i);
```
hwchen, line 74: Don't know how big the database is, but repeatedly calling `remove` could be expensive.

tesseract-server/src/handlers/metadata.rs
```diff
@@ -198,7 +207,38 @@ pub fn do_members(
...
+            ColumnData::NullableText(_) => Some(column.name.clone()),
+            _ => None,
+        })
+        .collect();
+
+    df.drain_filter(|datum, _| {
```
hwchen, line 222: It's a bit confusing to me that this param name is `datum` when it's a HashMap. And can the second param be removed?

tesseract-server/src/handlers/metadata.rs
```diff
@@ -198,7 +207,38 @@ pub fn do_members(
...
+        columns.iter().all(|name| {
```
hwchen, line 223: It looks like you're saying that you'll remove a row (`row` should probably be part of `drain_filter`'s name) if:
- for every text column
- the text value doesn't contain the `filter pattern`.

As a first feeling, I think I would have expected `any` instead of `all`. If I'm understanding correctly, `all` implies that all text columns would need to contain or not contain the `filter pattern` in order for something to happen to a row. I might be misunderstanding something here though.

hwchen: Also, I guess I would have expected that you would filter on one particular column?

hwchen:
Well, I guess I'm a bit tired; of course you can't use `Vec::drain_filter` directly on `Dataframe`.

Overall, I'd say that if your tests pass, it should be fine to merge. But I'd want to continue to have a conversation about whether `drain_filter` can be improved. And there are some parts around `drain_filter` and `Datum` that would benefit from clarification, in terms of documentation in the code.

frabarz:
Well, I have to admit most of the decisions I made were influenced by the compiler telling me I can't do that other thing I used to do in javascript, and following just the very basics of CS. I'm fully open to any suggestion.

So, let me explain the PR a bit more.
You're right on the final purpose: given the list of members returned by the DB as a DataFrame, I want to check row by row if the label (or the ID, if its type is Text) matches a user-provided string, and get only the rows with a match. The project that raised this need outputted ~3 million members.

Originally I tried to put it in the SQL statement, but I'd risk adding a SQL injection vulnerability, so I went with the DataFrame since is the only step between the SQL composition and the output formatting.
I tried to implement a row filtering method for the DataFrame, so I investigated a similar method on `std::Vec`, and found `Vec::drain_filter`. Since the specific case is for very long lists, I thought removing elements would be the way to go to reduce the memory use during the processing, and I definitively don't want to clone the list or its items. To check against all the elements on the same row, I create a `HashMap` where I put pairs `<column.name, column.column_data[i]>`, and pass it as a `datum` to the filter `Fn`.

In runtime, I make a vector with the names of the columns whose type is `ColumnData::Text` or `ColumnData::NullableText`, and use it to iterate on their values from the HashMap. I used `Iter::all` because if there's no text columns, the empty iterator returns `true`. If the filter `Fn` returns `false`, all items on that index on every column are removed. I left the second parameter `index` as a utility to follow conventions, but the compiler then forced me to put it in the closure so I used a `_`.

I've been thinking a bit more on the implementation, and I think it might be a better idea to do the filtering at the very last step, while generating the output in format.

hwchen:
It looks like we could put in place more pieces to protect against Sql injection. That might be a better way long-term.

For something like filters (and perhaps some other parts of tesseract, I should double-check that everything is parsed), we can do some validation with regex.

It looks like clickhouse now supports prepared statements, at least for some interfaces. I don't think the rust driver does yet, but this might be helpful eventually. https://clickhouse.tech/docs/en/whats-new/changelog/2019/#new-feature-6

And I think you already restrict privileges for the "public" user on the clickhouse instance.

As for the current implementation:

It's possible that filtering during the conversion to output would be more efficient, since you don't have to create another intermediate filtered dataframe.

The issue with just `remove`ing items, as I think you were doing in `drain_filter`, is that each removal should trigger a "shift" of the tail, which would be pretty expensive.

I guess I'm still not sure why it's necessary to check against all columns (or even all text rows), would you prefer to just check against one column?

frabarz:
I agree tesseract needs a SQL sanitizer, or a ORM, to interact with clickhouse. I also noticed the SQL for members is generated on `tesseract-core` instead of `tesseract-clickhouse`, is that not a problem?

I left the implementation open, thinking in "a DataFrame" more than in "a DataFrame of members", but I'm also considering that
- in some cases, members only have their `ID` and no `Label`
  - I'm not completely sure if the `ID` column is always numeric; if it also can be `Text` then the ID is what I want to check against
  - I could also say filtering by a numeric ID is worth it, an example would be a ZIP code
- the user can set a `locale` parameter, and for every locale code included, the DataFrame gets a new text column that might be worth to check

I was also thinking on how `Iter`s work after calling its methods, but that might complicate the DataFrame struct unnecessarily.

hwchen:
I don't think there will be an ORM for clickhouse any time soon. Personally, I'm comfortable with writing raw sql, although it would be nice for it to be more structured than what is currently in tesseract.

Generating sql for members in `core` has not been a problem so far... I think it's ok to leave it there until we know there will be a problem.

(As an aside, I prefer that zip codes be stored as strings)

I see that the requirements are a bit complicated for filtering. It actually makes me feel like that makes it even better to generate sql instead. It's not only more efficient, but I get the feeling that:
- Doing the filtering more broadly (like checking across all text cols) could result in unexpected behavior.
- Doing the filtering more specifically requires a lot of fancier type programming, especially if you want to specify multiple columns to filter against

PR #308, opened 2021-02-24
adds healthcheck endpoint

### LINKED ISSUE #307: Create a /healthcheck endpoint
Thomasnet is asking for a healthcheck endpoint to monitor the uptime of tesseract. It's a good idea to have it.
Healthcheck endpoints are usually called each minute, so the controller should just test light operations and live connections with other services, like the database. If everything is ok, return 200; if something's wrong, return 503.

frabarz:
This PR adds a `/healthcheck` endpoint to the tesseract-server crate, and the needed logic for it to work. This includes access to a ping function for backends. The implementation is done for clickhouse and mysql, postgres is still on hold because the way is implemented is a bit confusing for me.
This PR also includes a check in the e2e test case.

Closes #307


hwchen (approved):
Looks good. Just trying to think if there's another way to do timeouts besides pulling tokio in. Perhaps it doesn't really matter, since timeout handling will need to be changed with the updating of dependencies anyways (tokio moved to a somewhat different strategy for exposing timeout API)

PR #309, opened 2021-02-28
Chore/update deps

hwchen:
Updating dependencies.

Major changes include:
- clickhouse-rs to 1.0
- actix-web to 4.0
- async/await, futures to latest
- tokio to 1.0

### DIFF
--- a/tesseract-clickhouse/tests/integration.rs
+++ b/tesseract-clickhouse/tests/integration.rs
@@ -3,43 +3,26 @@ extern crate tokio;
 
 // Basic testing structured taken from
 // clickhouse-rs database tests by suharev7.
-use tokio::prelude::*;
-use std::{
-    env
-};
-use clickhouse_rs::{
-    types::Block, ClientHandle, Pool,
-};
-use clickhouse_rs::types::Complex;
+use std::env;
+use clickhouse_rs::Pool;
 
 fn database_url() -> String {
     let tmp = env::var("TESSERACT_DATABASE_URL").unwrap_or_else(|_| "tcp://localhost:9000?compression=lz4".into());
     tmp.replace("clickhouse://", "tcp://")
 }
 
-fn run<F, T, U>(future: F) -> Result<T, U>
-where
-    F: Future<Item = T, Error = U> + Send + 'static,
-    T: Send + 'static,
-    U: Send + 'static,
-{
-    let mut runtime = tokio::runtime::Runtime::new().unwrap();
-    let result = runtime.block_on(future);
-    runtime.shutdown_on_idle().wait().unwrap();
-    result
-}
-
-#[test]
-fn test_ping() {
+#[tokio::test]
+async fn test_ping() {
     // This test is meant as a sanity check
     // to ensure the docker provisioning process worked
     let pool = Pool::new(database_url());
-    let done = pool.get_handle().and_then(ClientHandle::ping).map(|_| ());
-    run(done).unwrap()
+    println!("{:?}", pool);
+    let mut client = pool.get_handle().await.unwrap();
+    client.ping().await.unwrap();
 }
 
-#[test]
-fn test_query() {
+#[tokio::test]
+async fn test_query() {
     #[derive(Debug, Clone, PartialEq)]
     pub struct RowResult {
         pub month_name: String,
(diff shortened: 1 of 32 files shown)


tesseract-clickhouse/tests/integration.rs
```diff
@@ -49,29 +32,18 @@ fn test_query() {
     // to ensure the SQL ingestion worked
     let pool = Pool::new(database_url());
     let sql = "SELECT month_name FROM tesseract_webshop_time;";
```
hwchen, line 34: @frabarz wanted to ask about this table in the test. It seems to be hardcoded, without any initialization. I'm not quite sure why the test needed here, seems like it's testing something very specific in a workflow.

hwchen:
@frabarz @alexandersimoes I came to a good stopping point to show some of the work.

I had started from the outside in for a bit (working on tesseract-server), but came to the point where I needed to go inside-out.

After the last two commits, `core` and `clickhouse` crates have been updated. You can see some of the types of changes I've made:

- using async/await syntax
- using `async-trait` for the `Backend` trait
- updating error crates from `failure` to `anyhow`
- silencing the `exec_sql_stream` method for now, since there's been substantial changes to `futures::Stream`
- updating `clickhouse` tests to use `tokio::test` instead of a custom runner.

hwchen:
btw, if you want to check out the changes to just `clickhouse` and `core` crates, you can run just the tests/compilation on those crates by running `cargo` from inside the directory, or specifying the crate in the workspace using command line args. That way you won't get all the errors from `server`.

hwchen:
Ok, postgres and mysql backends updated. I think that tomorrow I'll try to update everything in the server except for logic layer.

alexandersimoes:
Thanks for all the updates @hwchen!

I tried running the tests like you recommended and running `cargo test` in the `tesseract-core` dir worked great, but I received a failed test for `test test_query` in the `tesseract-clickhouse` dir.

Here's a stacktrace:

Screen Shot 2021-03-09 at 8 23 27 PM

hwchen:
That test is something I asked @frabarz about in another comment on the PR. I’d like to figure out if it’s specific to some deployment

frabarz:
Sorry, didn't see the past comment.

AFAIK, Jonathan wrote these tests. These are configured to run against a clickhouse database, stored in a docker container with a copy of the https://github.com/tesseract-olap/tesseract-example-app repository.

It's just for CI. The [travis config](https://github.com/tesseract-olap/tesseract/blob/master/.travis.yml) has the detail of the process in the `before_install` key, but to summarize, it runs the [./docker-compose.yml](https://github.com/tesseract-olap/tesseract/blob/master/docker-compose.yml) recipe, which in turn pulls the `yandex/clickhouse-server:19.3` image to create a container, and runs the [./docker/scripts/init_db.sh](https://github.com/tesseract-olap/tesseract/blob/master/docker/scripts/init_db.sh) script inside it.

alexandersimoes:
Ah ok, I didn't see that other comment.

hwchen:
Chipping away. Was wondering, does any instance of tesseract use the streaming option? (which allows streaming of http responses). If not, I'm going to comment it out for a bit and deal with it later.

alexandersimoes:
For now we are not using this release candidate anywhere anyway, so there is no problem with commenting out the streaming option for now.

hwchen:
`tesseract-server` compiles! There's some redis and logic-layer stuff that shouldn't work right now, and there are probably other bugs. But you can run the executable now. I don't have clickhouse/schema all set up, so if you try it out let me know what doesn't work.

My plan for next week is to bring back logic layer, then maybe redis. If I can get setup with a test environment I'll prioritize that first.

alexandersimoes:
Great news @hwchen! Do you need anything from me to get setup with a test env? Otherwise I'll see if someone @ dw can run the executable from this branch.

hwchen:
Just need a clickhouse instance with some data, and a corresponding tesseract schema. It's probably simplest if I just run tesseract locally against a remote clickhouse, but I'm also happy to do everything on a vm with ssh access if you prefer.

alexandersimoes:
Great, I'll coordinate with you in an email for the connection details.

hwchen:
I've begun manual testing of tesseract basics.

I was able to connect to clickhouse with oec-tesseract schema, and run an aggregation no problem. I was also able to get schema/cube data. Haven't tried `members`, `flush`, or other endpoints.

To me, this indicates that most of the basic plumbing just works. I didn't change any fn logic that wasn't related to async/await, so I think that sql generation is probably just working.

I'll be following up, I have some questions about validating against the cube cache.

I'll also be creating a basic testing checklist to go down. Now that I know that the basics work, I'll do that next week.

And just to give a general idea of time, I've spent about 10-11 hours so far.

hwchen:
note: in this line, it looks like you intend to set `readonly` as a default. However, it's causing some issue with the clickhouse server, that `setting 1` is not recognized. For my manual testing, I set it to `None` instead of `Some(1)`. I'll look into how to fix this.

tesseract-server/src/handlers/aggregate.rs
```diff
@@ -97,47 +98,46 @@ pub fn do_aggregate(
     // this is in braces to explicitly the scope in which
     // req is borrowed, since req is moved later in the `map_err`
     {
-        let cache = req.state().cache.read().unwrap();
+        let cache = state.cache.read().unwrap();
         let cube_cache = some_or_404!(cache.find_cube_info(&cube), format!("Cube {} not found", cube));
```
hwchen, line 102: Noting that this validation caused complete failure of request when I wanted to run without the cache (it was taking a very long time). Is this the desired behavior?

hwchen:
A thought about cache start-up times. Is this ever an issue? (Is this with what the redis integration was for?).

It looks like total cache build takes 146 seconds, with many queries taking about 0.09 seconds. This means about 1600 queries, and there's only 222 tables.

Looks like there's some repeats, like for year:

```
 INFO  clickhouse_rs::types::query_result > [send query] select distinct year from complexity_eci_a_hs02_hs6
 INFO  tesseract_clickhouse               > Time for sql execution: 0.087
 INFO  clickhouse_rs                      > [ping]
 INFO  clickhouse_rs                      > [pong]
 INFO  clickhouse_rs::types::query_result > [send query] select distinct year from complexity_eci_a_hs02_hs6
 INFO  tesseract_clickhouse               > Time for sql execution: 0.087
```

hwchen:
I'm actually not sure if it's because of shared dimensions. It doesn't look like what you're describing in your last sentence, because these duplicates are all showing up right next to each other. I had figured that it was something in the structure of how a schema was written.

alexandersimoes:
Ok good to know, we should keep an eye on it

